PR #26, opened 2015-03-02
Withfpu

bandvig:
Well. As the good state of FPU is achieved, it is time for pull request.
To use FPU:
(1) add all files from verilog/pfpu32 folder into your project
(2) add .FEATURE_FPU("ENABLED") in parameter list of mor1kx instance

Comparison with OR1200's FPU.

Operation clock cost for:  ADD/SUB - MUL - DIV - I2F/F2I
Port of OR1200 FPU     8  -   36 -  35 -    6
Pipelined FPU          5  -    6 -  18 -    3

Size of standalone FPU (LUTs as logic, Spartan6, ISE14.6):

Port of OR1200 FPU      : 2966 /  96 MHz
Pipelined FPU
  DSP blocks arn't used : 2822 / 112 MHz
  DSP blocks are used   : 1554 /  76 MHz  (+5 DSP modules)

Performance gain by Whetstone benchmark is up to 2 ... 3.5 times against OR1200's FPU and up to 20 times against soft-float depending on test. Detailed comparison will be published in mailing list after OpenRISC.NET operation restoring.

Additionally there is an extension to FPCSR. Thanks to the extension the each of FPU's exceptions  could be masked separately. The extension is blocked by default till approve from community. To activate the extension, uncomment the OR1K_FPCSR_MASK_FLAGS macro in mor1kx-sprs.v


rtl/verilog/mor1kx_lsu_cappuccino.v
```diff
@@ -659,8 +659,8 @@ endgenerate
 		   (exec_op_lsu_load_i | exec_op_lsu_store_i) ?
 		   exec_lsu_adr_i : ctrl_lsu_adr_i;
    assign dc_adr_match = dmmu_enable_i ?
-			 {dmmu_phys_addr[OPTION_OPERAND_WIDTH-1:2],2'b0} :
-			 {ctrl_lsu_adr_i[OPTION_OPERAND_WIDTH-1:2],2'b0};
+			 dmmu_phys_addr : //{dmmu_phys_addr[OPTION_OPERAND_WIDTH-1:2],2'b0} :
+			 ctrl_lsu_adr_i;  //{ctrl_lsu_adr_i[OPTION_OPERAND_WIDTH-1:2],2'b0};
```
skristiansson: This change doesn't seem to be related to the FPU?

rtl/verilog/mor1kx_execute_alu.v
```diff
@@ -669,6 +726,25 @@ endgenerate
      endcase // case (opc_alu_secondary_i)
 
    //
+   // Comparison logic
+   // To update SR[F] either from integer or float point comparision
+   generate
+     /* verilator lint_off WIDTH */
+     if (FEATURE_FPU!="NONE") begin
+     /* verilator lint_on WIDTH */
+       assign flag_set_o   = fpu_op_is_cmp ?
```
skristiansson: I think it'd be nicer to just let fpu_op_is_cmp evaluate to zero when FEATURE_FPU=="NONE" and get rid
of the generate here.
That way if the flag_set_o logic would happen to change, there is only one code path to change.
The tools should be able to optimize the logic away from that.

rtl/verilog/mor1kx_execute_alu.v
```diff
@@ -703,6 +779,21 @@ endgenerate
    assign op_cmov = op_alu_i & opc_alu_i == `OR1K_ALU_OPC_CMOV;
 
    // Result muxing - result is registered in RF
+generate
+/* verilator lint_off WIDTH */
+if (FEATURE_FPU!="NONE") begin
+/* verilator lint_on WIDTH */
+   assign alu_result_o = 
+     ({OPTION_OPERAND_WIDTH{op_logic}} & logic_result) |
```
skristiansson: Is the change from 'op_xxx ?' to '({OPTION_OPERAND_WIDTH{op_xxx}' motivated by something?
I've been experimenting with this before without any clear result that it'd be using less resources,
and I think 'op_xxx ?' is a clearer way to express it.

Also see the comment above about dropping the generate here and let fpu_arith_valid be tied to 0.

rtl/verilog/mor1kx_execute_alu.v
```diff
@@ -734,10 +827,24 @@ endgenerate
...
+/* verilator lint_off WIDTH */
+if (FEATURE_FPU!="NONE") begin
+/* verilator lint_on WIDTH */
+   assign alu_stall = 
+          (op_div_i & (~div_valid)) |
+		      (op_mul_i & (~mul_valid)) |
+          (fpu_op_is_arith & (~fpu_arith_valid)) |
+          (fpu_op_is_cmp & (~fpu_cmp_valid)) |
+		      (op_shift_i & (~shift_valid)) |
+		      (op_ffl1_i & (~ffl1_valid));
```
skristiansson: Indentation is odd here.
But the comment about dropping generate applies here too.

rtl/verilog/mor1kx_ctrl_cappuccino.v
```diff
@@ -1387,7 +1487,7 @@ end
 endgenerate
 
 generate
-if (FEATURE_FPU!="NONE") begin : fpu_ctrl
+if (FEATURE_FPU!="NONE") begin
```
skristiansson: Why was the label removed here?
IIRC some tools doesn't like unlabeled generate statements.

rtl/verilog/mor1kx_execute_alu.v
bandvig: You are right. Moreover, fpu_op_is_cmp and fpu_cmp_valid are alredy set to zero earlier in generate around FPU instance. So, the generate is redundant. I propose to make modification in a commit after merge.

rtl/verilog/mor1kx_execute_alu.v
bandvig: Oh! Indentation is odd because of difference in "tab" size, I think. I believe it is a good time to add in todo-list the action "replace tabs by spaces everywhere". And agree about generate. Again, I think, it would be nicer to make he modifications after merge into master. I'm ready to get the action item on myself.

rtl/verilog/mor1kx_execute_alu.v
bandvig: I achieved small benefit with replacing conventional multiplexer code by this-like one in several places in pfpu32_rnd.v. I checked that with synthesis standalone FPU unit with ISE14.6 for Spartan6. And the multiplexer is much longer than in pfpu32_rnd. So, that was motivation.
Agree about generate again.

rtl/verilog/mor1kx_lsu_cappuccino.v
bandvig: Well, let me describe the reason in more details than I did yesterday in IRC channel. 
(1) I wrote a very simple SoC for simulation proposes only. The SoC includes CPU, SRAM (not synthesizable but with preloading ability), BOOTROM, UART and INTERCON from MinSOC. I tried to run “Home World” (compiled with NewLIB based tool chain) and I found that It wasn’t work due to It was not possible to initialize the UART. In next turn the reason was that UART registers weren’t accessible due to zero the bits. 
(2) on the other hand the U-boot itself operates correctly on the real board. And any program loaded by U-boot operates correctly. For me it means that U-boot use a kind of another CPU and board initialization sequences than standalone NewLIB application. 
(3) Additionally, I’have found that during UART initialization performed by NewLIB application the LSU state machine is in WRITE state and access to bus goes through store_buffer_radr. However! In subsequent access the uart_putc tries to read line status (before put a symbol) and the access to line status register goes through IDLE state were ctrl_lsu_adr_i is used directly without forcing zeroes in the two LSBs.
Conclusion 
Perhaps, the CPU and Board initialization sequence in libgloss is not accurate enough. 
Of course, we could revert the commit either before or after merging.

rtl/verilog/mor1kx_ctrl_cappuccino.v
bandvig: During FPU debug I deep dived into synthesis outputs trying to find FPU related warnings. And long-long paths with included labels drove me mad. So, at some point I removed all FPU related labels including those ones I made myself. Of course, in the scope of your notice the labels have to be restored. I agree.

bandvig:
Well, what about merging? I see the two possible ways. 
The 1-st one. Merge all the modifications as is. Then I'll perform series of corrections in the master branch. The corrections will include: (a) remove extra generate and restoring old style of ALU’s output multiplexer, (b) restore label in all generate sections, (c) roll back forced zeroes in LSU generated addresses (I’m going to convert it to issue for mor11kx or newlib)
The 2nd one.  The back order. Close the pull request. Make all mentioned modifications in “withfpu” branch. Create new pull request.
What would be better? 

@skristiansson

skristiansson:
I'd prefer we do the corrections before merging, and as I also mentioned on IRC,
it'd be nicer to just apply this as a single (or a couple) commit(s),
without the intermediate steps leading up to the current state (and without the back merges).
If you could do that I agree that this is all ready to be merged.

bandvig:
@skristiansson 
Stefan, as I’m not very familiar with Git, I do not clearly understand how to implement commit without intermediate steps. To create this pull request I just used button on GitHub page. Could you give me a short guide?

skristiansson:
Easiest way to squash your commits into a single commit would be (assuming you are on withfpu branch and we create a new branch for the new pull request):
git fetch origin
git checkout -b fpu-pr
git reset --soft origin/master
git commit

bandvig:
Close the request to create another one with lesser nombers of cmmits

PR #44, opened 2016-10-04
Fixes for unused wires and witdh mismatches

olofk:
This silences a bunch of warning from the EDA tools


rtl/verilog/mor1kx_store_buffer.v
```diff
@@ -61,7 +61,7 @@ module mor1kx_store_buffer
      if (rst)
        write_pointer <= 0;
      else if (write_i)
-       write_pointer <= write_pointer + 1;
+       write_pointer <= write_pointer + 1'b1;
```
skristiansson: Bikeshedding, but I think I prefer the style you used in mor1kx_execute_alu
I.e. 1'd1

rtl/verilog/mor1kx_cfgrs.v
```diff
@@ -32,7 +32,7 @@ module mor1kx_cfgrs
     parameter OPTION_DCACHE_SET_WIDTH    = 9,
     parameter OPTION_DCACHE_WAYS         = 2,
     parameter FEATURE_DMMU               = "NONE",
-    parameter OPTION_DMMU_SET_WIDTH      = 6,
+    parameter [2:0] OPTION_DMMU_SET_WIDTH      = 6,
```
skristiansson: Why is this only needed for OPTION_DMMU_SET_WIDTH?
and not for:
OPTION_DMMU_WAYS
OPTION_IMMU_SET_WIDTH
OPTION_IMMU_WAYS
OPTION_DCACHE_SET_WIDTH
OPTION_ICACHE_SET_WIDTH
as well?

olofk: They are needed too. Just never finished catching all errors. Would you prefer to have sized parameters (as above) or picking a slice when doing the assignment to the SPRs?

rtl/verilog/mor1kx_store_buffer.v
olofk: Fixed in a follow-up patch. Also added read_pointer that I had missed before

rtl/verilog/mor1kx_cfgrs.v
wallento: I am wondering if that is really needed. Which tool complained about this?

olofk:
Quartus complains. Some examples

    Warning (10230): Verilog HDL assignment warning at mor1kx_cfgrs.v(99): truncated value with size 32 to match size of target (4) File: /path/to/mor1kx_5.0/rtl/verilog/mor1kx_cfgrs.v Line: 99
    Warning (10230): Verilog HDL assignment warning at mor1kx_cfgrs.v(148): truncated value with size 32 to match size of target (3) File: /path/to/mor1kx_5.0/rtl/verilog/mor1kx_cfgrs.v Line: 148

wallento:
The problem I see with setting the parameter width is that it won't work for `OPTION_DMMU_WAYS` because it uses `OPTION_DMMU_WAYS-1`, right? It needs `[2:0]` to represent the maximum value to put into the `[1:0]`. Can you please check for me how Quartus will be happy for that one?

stffrdhrn:
Closing this as @wallento merged with the Linter fixes PR #51

PR #54, opened 2017-10-16
Make PCU work at all, and then make it work correctly

### LINKED ISSUE #46: PCU: misses events
The PCU will at most increment the counters by one in a cycle. If there is more than one event enabled for a register, they are not counted. The workaround is to enable at most one event class per counter.

whitequark:
See individual commits. Right now it's completely useless.

Fixes #46.


wallento:
lgtm, ready to merge

PR #57, opened 2017-12-13
Fix compiling on modelsim after iverilog fixes

stffrdhrn:
The changes to get mor1kx compiling on iverilog seems to have broke
modelsim:

  External function 'mor1kx_rf_cappuccino.calc_rf_addr_width' may not be
  used in a constant expression.

To fix this (hopefully once and for all), I move the function out to the
mor1kx_utils.vh instead of calling it via internal references.


olofk:
Tested simulating with Icarus, modelsim and verilator and built with Quartus and ISE and everything seems fine. Thanks for the patch and sorry for the delay. Picked and pushed.

Issue #69, opened 2019-01-10
Building with default parameters fails.


stffrdhrn:
Tested with [de0_nano 1.1](https://github.com/olofk/de0_nano), but pointed it to mor1kx master.

It fails with the following error.

```
Info (12128): Elaborating entity "mor1kx_rf_cappuccino" for hierarchy "de0_nano_core:core|mor1kx:mor1kx0|mor1kx_cpu:mor1kx_cpu|mor1kx_cpu_cappuccino:cappuccino.mor1kx_cpu|mor1kx_rf_cappuccino:mor1kx_rf_cappuccino"
Error (10198): Verilog HDL error at mor1kx_rf_cappuccino.v(275): part-select direction is opposite from prefix index direction File: /home/shorne/work/openrisc/fuse-builds/de0_nano_1.1/src/mor1kx_5.0-r3/rtl/verilog/mor1kx_rf_cappuccino.v Line: 275
Info (10499): HDL info at mor1kx_rf_cappuccino.v(275): see declaration for object "wb_rfd_adr_expand"
Error (12152): Can't elaborate user hierarchy "de0_nano_core:core|mor1kx:mor1kx0|mor1kx_cpu:mor1kx_cpu|mor1kx_cpu_cappuccino:cappuccino.mor1kx_cpu|mor1kx_rf_cappuccino:mor1kx_rf_cappuccino" File: /home/shorne/work/openrisc/fuse-builds/de0_nano_1.1/src/mor1kx_5.0-r3/rtl/verilog/mor1kx_cpu_cappuccino.v Line: 1131
Info (144001): Generated suppressed messages file /home/shorne/work/openrisc/fuse-builds/de0_nano_1.1/synth-quartus/de0_nano_1_1.map.smsg
Error: Quartus II 64-Bit Analysis & Synthesis was unsuccessful. 2 errors, 68 warnings
    Error: Peak virtual memory: 1283 megabytes
    Error: Processing ended: Wed Jan  9 22:33:16 2019
    Error: Elapsed time: 00:01:45
    Error: Total CPU time (on all processors): 00:00:20
```

stffrdhrn:
It should be fixed now

PR #70, opened 2019-01-11
Utils: Fix zero as a shadow gpr parameter

stffrdhrn:
For the  mor1kx defaults the shadow GPR's is set to zero, this is used
when calculating the total register file address size but zero is not
supported.  Fix that by supporting 0.

Also, there is an issue with the expand wire assign when there are no
SHADOW SPRs, don't assign in that case.

This fixes github issue #69.


stffrdhrn:
thank you

PR #72, opened 2019-01-25
Marocchino devel

stffrdhrn:
This is a WIP to get marocchino working in the or1k-tests suite with mor1kx-generic.  You can see my repos for those.  This compiles now, but still some work to do.

Comments:
  - During this integration I found marocchino is almost a complete rewrite of mor1kx.  Should we just separate it out to a new repo i.e. openrisc/marocchino?  This will give it better exposure, and I am not so sure integrating it into mor1kx main will give it the credit it needs.
  - The main issues I see with integrating it is:
     1. It uses a different wb interconnect (i.e. so there are new signals from WB to the pipeline)
     2. It doesnt use the toplevel mor1kx module

### DIFF
--- a/bench/verilog/mor1kx_monitor.v
+++ b/bench/verilog/mor1kx_monitor.v
@@ -25,15 +25,14 @@
 `ifndef CPU_WRAPPER
  `define CPU_WRAPPER `MOR1KX_INST.mor1kx_cpu
 `endif
-`define CPU_INST `CPU_WRAPPER.`MOR1KX_CPU_PIPELINE.mor1kx_cpu
 `define EXECUTE_STAGE_INSN `CPU_WRAPPER.monitor_execute_insn
 `define EXECUTE_STAGE_ADV `CPU_WRAPPER.monitor_execute_advance
 `define CPU_clk `CPU_WRAPPER.monitor_clk
 `define CPU_FLAG `CPU_WRAPPER.monitor_flag
 `define CPU_SR `CPU_WRAPPER.monitor_spr_sr
 `define EXECUTE_PC `CPU_WRAPPER.monitor_execute_pc
-`define GPR_GET(x) `CPU_INST.get_gpr(x)
-`define GPR_SET(x, y) `CPU_INST.set_gpr(x, y)
+`define GPR_GET(x) `CPU_WRAPPER.get_gpr(x)
+`define GPR_SET(x, y) `CPU_WRAPPER.set_gpr(x, y)
 
 `include "mor1kx-defines.v"
 
--- a/mor1kx.core
+++ b/mor1kx.core
@@ -27,6 +27,7 @@ filesets:
       - rtl/verilog/mor1kx_decode_execute_cappuccino.v
       - rtl/verilog/mor1kx_decode.v
       - rtl/verilog/mor1kx_dmmu.v
+      - rtl/verilog/mor1kx_dpram_en_w1st.v
       - rtl/verilog/mor1kx_execute_alu.v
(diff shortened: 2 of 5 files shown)


bandvig:
@stffrdhrn 

> * During this integration I found marocchino is almost a complete rewrite of mor1kx.  

You are right. MAROCCHINO is almost complete rewrite of CAPPUCCINO (if be more precise). I did that to exclude extra variability caused by large number of OPTION_*  parameters and to separate my additional modifications (like pseudo CDC and ORFPX64A32) from other sources. I wished to prevent of exponential growing configuration-related “if”s.

>Should we just separate it out to a new repo i.e. openrisc/marocchino?  This will give it better exposure, and I am not so sure integrating it into mor1kx main will give it the credit it needs.

I haven’t got any objection. Perhaps we should ask other members of OpenRISC-team. Unfortunately, I’ve a problem with subscription for mailing list in LibreCores right now.  No problem if you rise the topic in the mailing list till I restore my subscription. The only thing. I would prefer “openrisc/mor1kx_marocchino” name for such repo.

> * The main issues I see with integrating it is:   
>   1. It uses a different wb interconnect (i.e. so there are new signals from WB to the pipeline)
>   2. It doesnt use the toplevel mor1kx module

I had to redesign Wishbone bridges and their interface with implementing pseudo-CDC. I moved part of atomic and snoop support logic into data-bridge in Wishbone clock domain.

stffrdhrn:
@bandvig 
 
> > Should we just separate it out to a new repo i.e. openrisc/marocchino?  This will give it better exposure, and I am not so sure integrating it into mor1kx main will give it the credit it needs.
> 
> I haven’t got any objection. Perhaps we should ask other members of OpenRISC-team. Unfortunately, I’ve a problem with subscription for mailing list in LibreCores right now. No problem if you rise the topic in the mailing list till I restore my subscription. The only thing. I would prefer “openrisc/mor1kx_marocchino” name for such repo.

Sure, since its derived from mor1kx that's fair.  I do believe the alu is still shared.  Copying/branching that should be fine though.

Let me forward this conversation to the list.

> > * The main issues I see with integrating it is:
> >   
> >   1. It uses a different wb interconnect (i.e. so there are new signals from WB to the pipeline)
> >   2. It doesnt use the toplevel mor1kx module
> 
> I had to redesign Wishbone bridges and their interface with implementing pseudo-CDC. I moved part of atomic and snoop support logic into data-bridge in Wishbone clock domain.

Yes, I noticed those signals.  It seems reasonable and makes your WB interface more smart/efficient.

bandvig:
@stffrdhrn 

> >  I would prefer “openrisc/mor1kx_marocchino” name for such repo.
> 
> Sure, since its derived from mor1kx that's fair.

Yep, and another minor reason. With such name the MAROCCHINO repo will be placed just before or after of “openrisc/mor1kx” in list of OpenRISC organization's repos :).


> I do believe the alu is still shared. Copying/branching that should be fine though.

No, ALU isn’t shared. The mor1kx_execute_alu module (in mor1kx_execute_alu.v file) is shared across CAPPUCCINO,  ESPRESSO an PRONTOESPRESSO. It includes circuits for divider, multiplier and all other operations. Opposite to that the mor1kx_execute_marocchino.v file includes three different modules: 

- pipelined multiplier (module mor1kx_multiplier_marocchino). By the way it doesn’t generate overflow similar to “PIPELINED” OpenRISC’s multiplier

- serial divider (module mor1kx_divider_marocchino)

-  and all other operations are placed into module mor1kx_exec_1clk_marocchino as these operations are 1-clock length

I know that good style is “one module per one file”. However, following that could be quite annoying sometimes :). 

> Let me forward this conversation to the list.

Sure.

PR #73, opened 2019-02-04
Marocchino: fixup monitoring to have insn inline with data

stffrdhrn:
Previously monitoring was broken because writeback data was not
available when we presented the instrucion to the monitor.  This
meant that the monitor always reported the wrong register values
and result values.

This patch adds some registers to register the insn's and pc's
when then insn is submitted to the order manager, the registers
are then read during writeback.

### DIFF
--- a/rtl/verilog/mor1kx_cpu_marocchino.v
+++ b/rtl/verilog/mor1kx_cpu_marocchino.v
@@ -564,8 +564,11 @@ module mor1kx_cpu_marocchino
   reg  wrbk_an_except_r;
 
   // Bench monitoring
+  localparam MONITOR_INSN_MEM_WIDTH = `OR1K_INSN_WIDTH + OPTION_OPERAND_WIDTH;
+  localparam MONITOR_NUM_EXTADR     = (1 << DEST_EXTADR_WIDTH);
+
   wire     [`OR1K_INSN_WIDTH-1:0] monitor_execute_insn/* verilator public */;
-  wire                            monitor_execute_advance/* verilator public */;
+  reg                             monitor_execute_advance/* verilator public */;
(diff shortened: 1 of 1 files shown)


stffrdhrn:
This fixes the monitor to properly work.  This is the output running a test in iverlog with marocchino

```
S 00000100: 18000000 l.movhi r0,0x0000       r0         = 00000000  flag: 0
S 00000104: a8200001 l.ori   r1,r0,0x0001    r1         = 00000001  flag: 0
S 00000108: c0000811 l.mtspr r0,r1,0x0011    SPR[0011]  = 00000001  flag: 0
S 0000010c: c1400000 l.mtspr r0,r0,0x5000    SPR[5000]  = 00000000  flag: 0
S 00000110: 18800000 l.movhi r4,0x0000       r4         = 00000000  flag: 0
S 00000114: a8840120 l.ori   r4,r4,0x0120    r4         = 00000120  flag: 0
S 00000118: 44002000 l.jr    r4                                     flag: 0
S 0000011c: 15000000 l.nop   0x0000                                 flag: 0
S 00000120: 04000254 l.jal   0x0000254                              flag: 0
S 00000124: 15000000 l.nop   0x0000                                 flag: 0
S 00000a70: b4600001 l.mfspr r3,r0,0x0001    r3         = 0000055f  flag: 0
...
```

stffrdhrn:
Thanks for the review, I pulled the old code from the other file in mor1kx_*_cappuccino.v thats why there was some inconsistency in spacing.  I didnt realize marocchino didnt use tabs.  But its good, I agree with no tabs.

I will fix it.  This is something we will want to put in the readme (under code style) when splitting out Marocchino.

stffrdhrn:
I think it looks ok now.  Merging.

PR #77, opened 2019-03-06
Fusesoc CI

stffrdhrn:
This PR adds a icarus simulator continuous integration that runs or1k-tests for cappuccino and espresso.   If the tests fail the CI fails.

This also contains some fixes to espresso and cappuccino to get some of the tests to pass.  Espresso tests are all still not passing.

### DIFF
--- a/.travis/install-fusesoc.sh
+++ b/.travis/install-fusesoc.sh
@@ -1,0 +1,45 @@
+#!/bin/sh
+
+set -x
+
+mkdir -p $HOME/src/cores
+mkdir -p $HOME/src/tools
+mkdir -p $HOME/tools
+
+# Get our toolchain
+cd $HOME/src/tools
+curl --remote-name --location \
+  https://github.com/stffrdhrn/gcc/releases/download/or1k-9.0.0-20181113/or1k-elf-9.0.0-20181112.tar.xz
+tar xC $HOME/tools -f $HOME/src/tools/or1k-elf-9.0.0-20181112.tar.xz
+
+PATH="$HOME/tools/or1k-elf/bin:${PATH}"
+
+# Get iverilog latest source, the version in trusty is no good
+cd $HOME/src/tools
+curl --remote-name --location \
+  http://shorne.noip.me/downloads/verilog-10.2.tar.gz
+tar xf verilog-10.2.tar.gz
+cd verilog-10.2
+./configure
+make
+sudo make install
+
+export PATH
+
+cd $HOME/src/tools
+git clone https://github.com/openrisc/or1k-tests.git ;\
+cd or1k-tests/native
+make
+
+# Get the cores for our test to run (in the future these should be in fusesoc)
+cd $HOME/src/cores
+git clone https://github.com/stffrdhrn/mor1kx-generic.git
+git clone https://github.com/stffrdhrn/intgen.git
+
+# Setup fusesoc
+fusesoc init -y
+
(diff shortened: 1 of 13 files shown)


olofk:
ehm... this approved thing was meant for one patch, not the whole series

rtl/verilog/mor1kx_ctrl_espresso.v
```diff
@@ -469,7 +469,7 @@ module mor1kx_ctrl_espresso
      if (rst)
        take_exception <= 0;
      else
-       take_exception <= (exception_pending | exception_r | doing_rfe_r) &
+       take_exception <= (exception_pending | exception_r) &
```
olofk: Have no knowledge of this area of the core, but I trust you on this.

stffrdhrn:
> Looks good! Happily surprised to see that icarus supports functions for evaluating parameters. This was not the case last time I checked

Thanks, I didn't have any issue with icarus I recall.  But there seems to be an issue with parameters and ventilator.  It might be a fusesoc bug.  But, I didn't get to the bottom of it.

i.e.

If I pass `--pipeline CAPPUCCINO` to the fusesoc with sim=verilator I expected it would set -Gpipeline=CAPPUCCINO during compilation.  However I found fusesoc was sending `--pipeline CAPPUCCINO` to the `Vorpsoc_top` ventilator binary instead.

Sorry, for bringing it up here.  I just remembered.

.travis/install-fusesoc.sh
```diff
@@ -0,0 +1,45 @@
...
+sudo make install
+
+export PATH
+
+cd $HOME/src/tools
+git clone https://github.com/openrisc/or1k-tests.git ;\
+cd or1k-tests/native
+make
+
+# Get the cores for our test to run (in the future these should be in fusesoc)
```
olofk: This uses some deprecated syntax that will be remove. I think the rest of the file could be replaced with 
fusesoc init -y
fusesoc library add mor1kx-generic https://github.com/stffrdhrn/mor1kx-generic
fusesoc library add intgen https://github.com/stffrdhrn/intgen
I'm not sure why $TRAVIS_BUILD_DIR is added to fusesoc.conf though. Can you elaborate?

stffrdhrn: This sounds good.

> I'm not sure why $TRAVIS_BUILD_DIR is added to fusesoc.conf though. Can you elaborate?

$TRAVIS_BUILD_DIR is the root of `mor1kx` git repo, its the current git branch under test (i.e. the PR code) already checked out onto the travis build host.  We need fusesoc to add it to the core list.

olofk:
Re verilator: For CAPI1 cores, arguments to verilator gets passed straight to the model unless you set `cli_parser=fusesoc` in the verilator section. For CAPI2 this is the default behaviour

olofk:
I think this looks good. Hard to comment on the RTL changes, and there are some things that could be cleaned up (like install-fusesoc.sh doing more than just installing fusesoc), but I think it's better to land these changes and take it from there

.travis/install-fusesoc.sh
stffrdhrn: FYI, after running it the config will be something like

```
[main]
cores_root = /home/travis/src/cores
  /home/travis/build/mor1kx/master
```

Can we just do `fusesoc library add $TRAVIS_BUILD_DIR`?

stffrdhrn: thanks, Ill make those clean ups before merging, it should be easy.

PR #80, opened 2019-05-13
Implementation of unordered FP-comparison.

bandvig:
As discussed in (https://lists.librecores.org/pipermail/openrisc/2019-May/001995.html) the pull request has been created.  Colleges, please give some words for review.

### DIFF
--- a/rtl/verilog/mor1kx-defines.v
+++ b/rtl/verilog/mor1kx-defines.v
@@ -213,11 +213,15 @@
 // MSbit indicates FPU operation valid
 `define OR1K_FPUOP_WIDTH	8
 `define OR1K_FPUOP_SELECT 7:0
-// FPU unit from Usselman takes 5 cycles from decode, so 4 ex. cycles
-`define OR1K_FPUOP_CYCLES 3'd4
+// Select bits for Ordered/Unordered comparison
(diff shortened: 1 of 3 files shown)


rtl/verilog/mor1kx-defines.v
```diff
@@ -213,11 +213,15 @@
 // MSbit indicates FPU operation valid
 `define OR1K_FPUOP_WIDTH	8
 `define OR1K_FPUOP_SELECT 7:0
-// FPU unit from Usselman takes 5 cycles from decode, so 4 ex. cycles
-`define OR1K_FPUOP_CYCLES 3'd4
+// Select bits for Ordered/Unordered comparison
+`define OR1K_FPUOP_CMP_ANY_WIDTH 3
+`define OR1K_FPUOP_CMP_ANY_SELECT 2:0
+// Unordered comparison bit
+`define OR1K_FPUOP_CMP_URD_BIT 5
```
stffrdhrn: I would recommend swapping urd for unordered here and other places. Urd is not easy to read.

bandvig: Such replacement will make names lengthy. On the other hand, you are right. I'll do this during next days.

stffrdhrn:
This looks good to me.

Issue #81, opened 2019-05-24
Replace custom image by providing LibreCores CI Image 


Nancy-Chauhan:
This issue is in direction of Setting up Demo CI Flow for mor1kx Project. Currently mor1kx constantly evolving code is verified by [Travis Continous Integration](https://github.com/openrisc/mor1kx/blob/master/.travis.yml). One of the first aim of GSoC 2019 project is to use Librecores CI for Continous Integration of mor1kx. [Librecores CI](https://www.librecores.org/static/librecores-ci) is an approach/service to provide continuous integration to hardware projects hosted on [Librecores](https://www.librecores.org/static/) to improve user experience and reliability.

**Goal :** This will create shared tooling of [Librecores CI image](https://github.com/librecores/docker-images/tree/master/librecores-ci) with mor1kx project 
**Implementation :** With reference to issue https://github.com/librecores/docker-images/issues/10 , it will be possible to replace custom dockerfile image used in mor1kx project https://github.com/openrisc/mor1kx/tree/master/.travis with Librecores CI image . Further Jenkinsfile can be created for refactoring of existing travis CI 

**_Project Link_**: https://github.com/librecores/librecores-ci/ 

 
 

CC @oleg-nenashev @wallento @stffrdhrn
[_Link to doc proposal_ ](https://docs.google.com/document/d/1d0mIrLpU80IqB5oe4CvUz5H3bMR84tg0P99doWrovek/edit)

oleg-nenashev:
Please add description and link to other tasks in your GSoC project scope

PR #82, opened 2019-05-29
Run tests inside librecores CI docker image

Nancy-Chauhan:
Solves issue #81


oleg-nenashev:
@Nancy-Chauhan please use mentions to request reviews

oleg-nenashev:
@stffrdhrn Hi. Did you have a chance to configure CI for your fork?

stffrdhrn:
@oleg-nenashev
Do you mean with Travis? Yes, the link is here, I am not sure if it's public 
https://travis-ci.org/stffrdhrn/mor1kx/branches

I haven't tried anything on librecores-ci yet.

.travis.yml
```diff
@@ -16,12 +19,11 @@ addons:
     - python3-pip
 
 before_install:
-  - sudo pip3 install --upgrade pip setuptools
-  - sudo pip3 install --ignore-installed six
-  - sudo pip3 install fusesoc
+  -  docker pull librecores/librecores-ci-dev
```
oleg-nenashev: You are using wrong image. https://hub.docker.com/r/librecores/librecores-ci/ should be used

.travis.yml
```diff
@@ -3,6 +3,9 @@ cache: ccache
 sudo: true
 dist: trusty
 
+services:
+  - docker 
+
 addons:
```
oleg-nenashev: I doubt it is needed after switching to the Docker image

Nancy-Chauhan:
CC @oleg-nenashev @stffrdhrn Can you please have a look. This seems to work with librecores-ci-image , but fails for verliator job and ESPRESSO pipeline.

stffrdhrn:
Let's keep it in parallel. Instead of having the test.sh script that runs everything. Just use stages and run a different docker cmd and have different environment variables per run.

stffrdhrn:
@oleg-nenashev @Nancy-Chauhan 
In addition to the last comment.  I've reason we use stages it to have a fall fast pipeline.  If verilator checks fail we don't want to go onto the regression testing.  Just fail fast.

Also, note we have marked espresso as known to fail.

With the test.sh we can't do either of those.

Nancy-Chauhan:
@stffrdhrn @oleg-nenashev  Can you please check the recent changes
Travis log : https://travis-ci.com/Nancy-Chauhan/mor1kx/builds/114753627

stffrdhrn:
> @stffrdhrn @oleg-nenashev Can you please check the recent changes
> Travis log : https://travis-ci.com/Nancy-Chauhan/mor1kx/builds/114753627

This looks great! If you are done I can merge this.

.travis/Dockerfile
```diff
@@ -12,7 +12,7 @@
 #  'RUN .travis/install-verilator.sh' to this dockerfile to have an image
 #  with the full environment already installed.
 #
-FROM ubuntu:trusty
+FROM librecores/librecores-ci:latest
```
stffrdhrn: It we run librecores/librecores-ci image in our tests do we need this `.travis/Dockerfile` anymore?  Do we use this file?

Nancy-Chauhan:
@oleg-nenashev  @stffrdhrn  I think it is done . But Stafford is the verilator log fine to you  ? Because librecores-ci uses the recent version

.travis/Dockerfile
Nancy-Chauhan: @stffrdhrn This dockerfile does the same work what it used to do before i.e to simulate the travis environment only for testing purposes,this is not actually used in travis . And it works fine Librecores-ci base image

stffrdhrn: Yes, but we needed to simulate the Travis environment to have something consistent for running tests. Now the tests run in docker so it's consistent.  This is not needed. Anyway I can remove.

oleg-nenashev:
I was a bit late to the party. Thanks @Nancy-Chauhan, it is great to see CI running in the repo again !

stffrdhrn:
I just merged because it's working. We can always do more changes.  We probably also want this change on the or1k_marocchino core.

Nancy-Chauhan:
Yes @stffrdhrn, Can you please give more reference on or1k_marocchino core.

stffrdhrn:
https://github.com/openrisc/or1k_marocchino?files=1

If you see the Travis file you will notice it's very similar.

Nancy-Chauhan:
@oleg-nenashev @stffrdhrn Yes I see it is very similar. Will open PR soon.

Issue #83, opened 2019-06-12
[GSoC 2019] Create a Jenkinsfile for Librecores CI in mor1kx 


Nancy-Chauhan:
This issue will be in direction of setting up demo CI flow for mor1kx in Librecores-CI. Initially, Jenkinsfine can be created which uses the same flow as in Travis CI in mor1kx and later it can be extended to include :  
- Yosys synthesis which is required for monitoring resource usages.
- OpenOCD/GDB  for CPU pipeline debugging
 
CC @oleg-nenashev, @stffrdhrn

PR #88, opened 2019-07-05
Update Jenkinsfile in mor1kx

Nancy-Chauhan:
Updated the version of librecores/librecores-ci docker image. In reference to #83

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -137,5 +137,5 @@ pipeline {
 }
 
 void dockerrun() {
-    sh 'docker run --rm -v $(pwd):/src -e "JOB=$JOB" -e "SIM=$SIM" -e "PIPELINE=$PIPELINE" -e "EXPECTED_FAILURES=$EXPECTED_FAILURES" -e "EXTRA_CORE_ARGS=$EXTRA_CORE_ARGS" librecores/librecores-ci /src/.travis/test.sh'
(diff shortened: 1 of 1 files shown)


stffrdhrn:
In July 5th @olofk added jtag_vpi r3 which the librecores-ci build is picking up and failing on.  This is likely a problem with that new version. 

1. Quick fix, we can update the core file to require r2
2. Longer fix, fix the root cause

stffrdhrn:
In July 5th @olofk added jtag_vpi r3 which the librecores-ci build is picking up and failing on.  This is likely a problem with that new version. 

1. Quick fix, we can update the core file to require r2
2. Longer fix, fix the root cause

Nancy-Chauhan:
Thanks, @stffrdhrn. I am working to fix this

stffrdhrn:
I can't merge when there is a failure. It sould be fixed now. Can you try the build again?  If it's green I can merge

Nancy-Chauhan:
@stffrdhrn Actually @oleg-nenashev and I tried running it but there was a disk space problem, not enough for parallel OpenRISC runs and Docker on the same host.https://github.com/librecores/librecores-ci-jenkins-server/issues/64, which will be fixed most probably next week

Nancy-Chauhan:
@stffrdhrn  Now I rebuild it!  Works fine now

PR #90, opened 2019-07-31
Syntax fix for yosys

olofk:


bandvig:
@olofk 
Is the variant still visible by other EDA tools (I'm not very familiar with them)?

olofk:
Yes. Having the comment after the case statement is the recommended way so this should work for all tools

PR #93, opened 2019-08-20
Use OpenRISC pipeline builder from pipeline library

Nancy-Chauhan:
- [x] Use pipeline builder from https://github.com/librecores/librecores-pipeline-lib/pull/5
- [x] Add Yosys reporting to pipeline

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -3,7 +3,7 @@ openriscPipeline {
     yosysReport {
 		core 'mor1kx'
 		target 'synth'
-		logPath 'build/mor1kx_5.0-r3/synth-icestorm/yosys.log'
+		logPath 'build/mor1kx_*/synth-icestorm/yosys.log'
 	}
 
     job('verilator') {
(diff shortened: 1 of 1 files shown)


stffrdhrn:
so far it looks good

Nancy-Chauhan:
Hello, @stffrdhrn @oleg-nenashev changes done!  
With this, Yosys should run successfully, tested on my local instance

Jenkinsfile
```diff
@@ -1,4 +1,11 @@
 openriscPipeline {
+
+    yosysReport {
+		core 'mor1kx'
+		target 'synth'
+		logPath 'build/mor1kx_5.0-r3/synth-icestorm/yosys.log'
```
stffrdhrn: again here we can't keep the version in the path name.

Nancy-Chauhan: @stffrdhrn This should work now. @oleg-nenashev will be looking into https://ci.librecores.org/ stability maybe in a couple of weeks

PR #108, opened 2019-12-19
Fix spr_eear to save appropriate value for each exception

### LINKED ISSUE #99: [Bug] spr_eear not saving instruction address when illegal instruction exception
As in spec, 
EEAR register should save the instruction fetch VA when illegal instruction exception.

We should change code in mor1kx_ctrl_cappuccino.v as follow.

```verilog
    always @(posedge clk `OR_ASYNC_RST)
     if (rst)
       spr_eear <= {OPTION_OPERAND_WIDTH{1'b0}};
     else if (/*padv_ctrl & exception*/ exception_re)
       begin
	  if (except_ibus_err_i | except_itlb_miss_i | except_ipagefault_i | except_illegal_i)
	    spr_eear <= pc_ctrl_i;
	  else if (except_dbus_i | except_dtlb_miss_i | except_dpagefault_i | except_align_i)
	    spr_eear <= ctrl_lsu_adr_i;
       end

### PR DESCRIPTION
**Pull request by JaewonHur**
This will resolve #99

### DIFF
--- a/rtl/verilog/mor1kx_ctrl_cappuccino.v
+++ b/rtl/verilog/mor1kx_ctrl_cappuccino.v
@@ -833,9 +833,9 @@ module mor1kx_ctrl_cappuccino
        spr_eear <= {OPTION_OPERAND_WIDTH{1'b0}};
      else if (/*padv_ctrl & exception*/ exception_re)
        begin
-	  if (except_ibus_err_i | except_itlb_miss_i | except_ipagefault_i)
+	  if (except_ibus_err_i | except_itlb_miss_i | except_ipagefault_i | except_illegal_i)
 	    spr_eear <= pc_ctrl_i;
-	  else
+	  else if (except_dbus_i | except_dtlb_miss_i | except_dpagefault_i | except_align_i)
 	    spr_eear <= ctrl_lsu_adr_i;
        end
 

### CODE REVIEW

**Comment by stffrdhrn** [comment]
Thanks @JaewonHur , I can't verify this did right now but I'm working on it.

Sorry for delaying. I hope we can get all these fixes in.

File: rtl/verilog/mor1kx_ctrl_cappuccino.v [hunk 1]
```diff
@@ -833,9 +833,9 @@ module mor1kx_ctrl_cappuccino
        spr_eear <= {OPTION_OPERAND_WIDTH{1'b0}};
      else if (/*padv_ctrl & exception*/ exception_re)
        begin
-	  if (except_ibus_err_i | except_itlb_miss_i | except_ipagefault_i)
+	  if (except_ibus_err_i | except_itlb_miss_i | except_ipagefault_i | except_illegal_i)
 	    spr_eear <= pc_ctrl_i;
-	  else
+	  else if (except_dbus_i | except_dtlb_miss_i | except_dpagefault_i | except_align_i)
```
skristiansson, line 838: why is this needed?
either this list is covering all the cases that are not covered in the previous if and then it is superfluous,
or it does not contain all the cases that are not covered in the previous if and then it is wrong

stffrdhrn: I agree, with verilog we should always have an else case otherwise the logic is incomplete. Also, the reason I am looking at this is because I noticed during linux debugging with mor1kx the trap instructions are appearing to have incorrect EEAR set.  The trap case doesn't seem to be handler here, maybe its somewhere else?

stffrdhrn: Well, my comment is wrong here `l.trap` doesn't set eear as per spec.

PR #110, opened 2019-12-19
Fix mor1kx_decode to accept valid l.fl1 (l.ff1) instructions

### LINKED ISSUE #102: [Bug] l.fl1 instruction not decoded
@lifeasageek

l.fl1 instruction is not decoded correctly.
Any l.fl1 instruction can reproduce the error.

We can solve this with following change in mor1kx_decode.v

```verilog
  assign decode_op_ffl1_o = opc_insn == `OR1K_OPCODE_ALU &&
                        (decode_insn_i[9:8] == 2'b00 || decode_insn_i[9:8] == 2'b01) &&
                        opc_alu == `OR1K_ALU_OPC_FFL1;
```

JaewonHur:
This resolves #102

### DIFF
--- a/rtl/verilog/mor1kx_decode.v
+++ b/rtl/verilog/mor1kx_decode.v
@@ -260,6 +260,7 @@ module mor1kx_decode
 			      opc_insn == `OR1K_OPCODE_SHRTI;
 
    assign decode_op_ffl1_o = opc_insn == `OR1K_OPCODE_ALU &&
+                             (decode_insn_i[9:8] == 2'b00 || decode_insn_i[9:8] == 2'b01) &&
 			     opc_alu == `OR1K_ALU_OPC_FFL1;
 
    assign decode_op_movhi_o = opc_insn == `OR1K_OPCODE_MOVHI;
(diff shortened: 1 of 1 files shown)


rtl/verilog/mor1kx_decode.v
```diff
@@ -260,6 +260,7 @@ module mor1kx_decode
 			      opc_insn == `OR1K_OPCODE_SHRTI;
 
    assign decode_op_ffl1_o = opc_insn == `OR1K_OPCODE_ALU &&
+                             (decode_insn_i[9:8] == 2'b00 || decode_insn_i[9:8] == 2'b01) &&
```
skristiansson: What is the exact problem you are trying to solve with this change?
While the change is not wrong per se, it doesn't do what the commit message says.
l.fl1 and l.ff1 are decoded correctly as is, the only issue is that an instruction with decode_insn_i[9] == 1 will be
decoded as l.fl1/l.ff1 as well.

I am not against merging this change, but please change the commit message to more properly describe the problem it solves.

JaewonHur: I think it is trivial, but there were some cases when insns with decode_insn_i[9] == 1 were also decoded as l.fl1/l.ff1.

Actually, I changed mor1kx_decode.v somewhat more, I will request pull if you need.

skristiansson: Yeah, I'm just asking you to update the commit message, the code change is fine.
Something along the lines of:

_Fix mor1kx_decode to only accept valid l.fl1 and l.ff1 instructions

Both l.fl1 and l.ff1 have bit 9 set in the insruction as port of their op codes,
mor1kx_decode does not currently properly check this_

PR #112, opened 2019-12-19
subtract offset (0x400) when accessing shadow registers

### LINKED ISSUE #104: [Bug] mtspr shadow gpr read
@lifeasageek

When reading or writing shadow gprs through mfspr (mtspr),
it seems mor1kx_rf_cappuccino doesn't offset the register address.
(As the spec, SPR[1024+x] points to GPR_'x')

You can reproduce error through below instructions
```
18000000 l.movhi r0,0            
a8200001 l.ori   r1,r0,0x1       
c0000811 l.mtspr r0,r1,0x11      
15000000 l.nop   0               
c0010c0e l.mtspr r1,r1,0x40e 
```

Changing the code in mor1kx_rf_cappuccino.v as follow should solve this issue.
```verilog
  assign rf_wradr = wb_rf_wb_i ? wb_rfd_adr_expand : (spr_bus_addr_i[RF_ADDR_WIDTH-1:0] - `RF_ADDR_WIDTH-1'h400);
```

JaewonHur:
This resolves #104

### DIFF
--- a/rtl/verilog/mor1kx_rf_cappuccino.v
+++ b/rtl/verilog/mor1kx_rf_cappuccino.v
@@ -275,7 +275,7 @@ if (FEATURE_DEBUGUNIT!="NONE" || FEATURE_FASTCONTEXTS!="NONE" ||
    assign wb_rfd_adr_expand[OPTION_RF_ADDR_WIDTH-1:0] = wb_rfd_adr_i;
 
    assign rf_wren =  wb_rf_wb_i | spr_gpr_we;
-   assign rf_wradr = wb_rf_wb_i ? wb_rfd_adr_expand : spr_bus_addr_i[RF_ADDR_WIDTH-1:0];
+   assign rf_wradr = wb_rf_wb_i ? wb_rfd_adr_expand : (spr_bus_addr_i[RF_ADDR_WIDTH-1:0] - `RF_ADDR_WIDTH-1'h400);
    assign rf_wrdat = wb_rf_wb_i ? result_i : spr_bus_dat_i;
 
    // Zero-pad unused parts of vector


rtl/verilog/mor1kx_rf_cappuccino.v
```diff
@@ -275,7 +275,7 @@ if (FEATURE_DEBUGUNIT!="NONE" || FEATURE_FASTCONTEXTS!="NONE" ||
    assign wb_rfd_adr_expand[OPTION_RF_ADDR_WIDTH-1:0] = wb_rfd_adr_i;
 
    assign rf_wren =  wb_rf_wb_i | spr_gpr_we;
-   assign rf_wradr = wb_rf_wb_i ? wb_rfd_adr_expand : spr_bus_addr_i[RF_ADDR_WIDTH-1:0];
+   assign rf_wradr = wb_rf_wb_i ? wb_rfd_adr_expand : (spr_bus_addr_i[RF_ADDR_WIDTH-1:0] - `RF_ADDR_WIDTH-1'h400);
```
skristiansson: This is not correct, the 1024 offset is already handled at https://github.com/openrisc/mor1kx/blob/master/rtl/verilog/mor1kx_rf_cappuccino.v#L262

JaewonHur: It seems that mor1kx_rf_cappuccino line 262 only checks whether mtspr/mfspr access gpr or not. (reg group:0, reg number: 1024~1535)

Isn't it right to subtract 1024 from spr_access_addr when writing/reading actual gpr?

skristiansson: No, it's not correct to subtract 1024 from spr_access_addr, RF_ADDR_WIDTH only spans (at max) over the 512 shadow registers. i.e. it can only be 0-512, subtracting 1024 from that would yield negative addresses.

PR #114, opened 2020-04-17
Fix mor1kx_decode to accept valid l.fl1 (l.ff1) instructions (#110)

rkrajnc:
* Fix mor1kx_decode to only accept valid l.fl1 and l.ff1 instructions

Both l.fl1 and l.ff1 have bit 9 set in the insruction as port of their op codes,
mor1kx_decode does not currently properly check this


rkrajnc:
Please disregard, some half-asleep misclicks ...

PR #118, opened 2021-02-27
Convert Travis CI to GitHub Actions

stffrdhrn:
Work in progress on migrating to Github actions for reasons:
  - Travis CI is going away some time
  - Travis CI is broken because of change in elf-loader and old or1k-tests in librecores-ci image
  - The librecores CI doesn't run very conistently


stffrdhrn:
ok, pulling the trigger

PR #120, opened 2021-06-11
Formal

Harshitha172000:
Formal Verification of Mor1kx using yosys formal tools.
Assertions added to the source files can be tested using SBY files (Path: bench/formal).


bench/formal/Makefile
```diff
@@ -0,0 +1,29 @@
...
+# Copyright (C) 2021
+#    Stafford Horne <shorne@gmail.com>
+#
+##########################################################$
+
+TESTS := mor1kx_cache_lru \
+	mor1kx_simple_dpram_sclk \
+	mor1kx_true_dpram_sclk \
+	mor1kx_icache \
+        mor1kx_immu
```
stffrdhrn, line 18: Can you fix indentation here.

bench/formal/mor1kx_immu.sby
```diff
@@ -0,0 +1,18 @@
...
+depth 20
+
+[engines]
+smtbmc yices
+
+[script]
+
+read -formal -D IMMU mor1kx_immu.v
+read -formal mor1kx_true_dpram_sclk.v
+prep  -top mor1kx_immu
```
stffrdhrn, line 12: Can you remove the space here to be consistent?

rtl/verilog/mor1kx_cache_lru.v
```diff
@@ -277,5 +277,20 @@ module mor1kx_cache_lru(/*AUTOARG*/
       // lru_post = 4'b0010 is what we expected.
    end
 
+/*-----Formal Checking-------*/
+
+`ifdef FORMAL
+
+   always @(*) begin
+      assert (NUMWAYS > 1);
+      assume ($onehot0(access));
```
stffrdhrn, line 286: Should this assume change to an assert with integration tests? i.e. be `ASSUME`

rtl/verilog/mor1kx_icache.v
```diff
@@ -483,4 +485,102 @@ module mor1kx_icache
...
+
+      end
+   end
+
+   /*-------Cover-------*/
+   always @(posedge clk) begin
+   // Cache Read   --->  Trace 0
+   cover (read && cache_hit_o && f_past_valid);
+   // Cache Refill --->  Trace 1
+   cover (refill && refill_done_o && refill_hit && $changed(cpu_dat_o) && f_past_valid);
```
stffrdhrn, line 581: Can you fix indentation here?

rtl/verilog/mor1kx_branch_predictor_saturation_counter.v
```diff
@@ -94,4 +94,17 @@ module mor1kx_branch_predictor_saturation_counter
...
+
+`ifdef FORMAL
+   reg f_past_valid;
+   initial f_past_valid = 1'b0;
+   always @(posedge clk)
+      if (!f_past_valid)
+         assume (rst);
+   always @(*)
+      if (f_past_valid)
+         assert (brn_taken !== 1'bx);
```
stffrdhrn, line 108: Since `f_past_valid` is never set high will anything work??

This is internal state, could we assert an external state instead?

rtl/verilog/mor1kx_branch_predictor_simple.v
```diff
@@ -28,4 +28,17 @@ module mor1kx_branch_predictor_simple
...
+/*--------Formal Checking---------*/
+
+`ifdef FORMAL
+`ifdef S_BRN
+   wire f_not_unknown;
+   initial f_not_unknown = 1'b0;
+   assign f_not_unknown = (op_bf_i !== 1'bx && ^immjbr_upper_i !== 1'bx && op_bnf_i !== 1'bx);
+   always @(*) begin
+      if (f_not_unknown)
+         assert (predicted_flag_o !== 1'bx);
```
stffrdhrn, line 40: This is quite a different approach to how you test the other *branch_predictor* modules.  It also looks more correct.  Why did you do it this way for this one?

stffrdhrn (commented):
I was not able to finish the review of all files.  Please have a look at the comments and fix similar items across all modules.  We still need to think about:
  - Iterface files
  - Doing less internal signal verification
  - Doing less asserts on assign signals, it seems like not very useful
  - Top level verification i.e. `riscv-formal`

rtl/verilog/mor1kx_branch_predictor_simple.v
Harshitha172000: It just had one assign statement, so I felt this method would be better. I will try testing all branch_predictor modules in this approach.

rtl/verilog/mor1kx_branch_predictor_gshare.v
```diff
@@ -113,4 +113,29 @@ module mor1kx_branch_predictor_gshare
...
+      if ($past(prev_op_brcond_i) && $past(padv_decode_i) && f_past_valid)
+         assert ($past(brn_hist_reg[GSHARE_BITS_NUM - 1 : 0] << 1) == brn_hist_reg[GSHARE_BITS_NUM - 1 : 1]);
+   end
+
+   always @(posedge clk)
+      if (f_past_valid)
+         assert (brn_taken !== 1'bx);
+
+   always @(*)
+      assert (prev_idx < 2 ** GSHARE_BITS_NUM);
```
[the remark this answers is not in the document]
Harshitha172000: No, I shouldn't have asserted this internal variable.

rtl/verilog/mor1kx_branch_prediction.v
```diff
@@ -110,4 +110,56 @@ end else begin
...
+                 ^immjbr_upper_i !== 1'bz);
+         assert (^brn_pc_i !== 1'bx || ^brn_pc_i !== 1'bz);
+         assert (prev_op_brcond_i !== 1'bx ||
+                 prev_op_brcond_i !== 1'bz);
+         assert (prev_predicted_flag_i !== 1'bx ||
+                 prev_predicted_flag_i !== 1'bz);
+         assert (flag_i !== 1'bx || flag_i !== 1'bz);
+         assert (padv_decode_i !== 1'bx || padv_decode_i !== 1'bz);
+         assert (execute_bf_i !== 1'bx || execute_bf_i !== 1'bz);
+         assert (execute_bnf_i !== 1'bx || execute_bnf_i !== 1'bz);
```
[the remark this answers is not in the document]
Harshitha172000: I will remove these statements and follow the approach used in simple branch prediction, maybe I don't even need to declare `f_past_valid` if I follow that approach.

rtl/verilog/mor1kx_branch_prediction.v
```diff
@@ -110,4 +110,56 @@ end else begin
...
+   initial f_past_valid = 1'b0;
+
+   always @(posedge clk)
+      if (!f_past_valid)
+         assume (rst);
+
+   //Converting all input design assumptions to assertions.
+   //After reset all inputs are expected to have either 1 or 0 bit.
+   always @(posedge clk) begin
+      if ($past(rst) && f_past_valid) begin
```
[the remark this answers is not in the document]
Harshitha172000: I have missed setting `f_past_valid`, I will add it in these modules.

stffrdhrn:
https://chris.beams.io/posts/git-commit/#why-not-how

stffrdhrn:
Can you add this to our github actions CI?  I think we can start merging the changes after a few things are fixed.

 1. The tests all pass, I am getting a failure on lsu
 2. Some of the review comments are addressed
 3. Sorry its a bit late, but I think you need to clean up your git commit comments, read the article I posted. 
    * https://chris.beams.io/posts/git-commit/
    * Your commit descriptions are not up to our standards i.e. a recent commit `formal test branch` that doesnt explain what the change is.
    * You will need to rebase and reword your git commits
    * I recommend squash and fixup to reduce your commits rather than just adding new commits. i.e. after a code review you can just fixup/amend a commit rather than adding a commit that says "FIx as per PR comments"
    * Add a commit body explaining why the changes are needed

Harshitha172000:
LSU passes the formal test now.

I think assertions to cpu_cappuccino are not enough, and there is no point in asserting the same properties which
are added in the submodules. How should I proceed further with this top module testing?

Also, what should I add to git actions ci?

stffrdhrn:
> LSU passes the formal test now.
> 
> I think assertions to cpu_cappuccino are not enough, and there is no point in asserting the same properties which
> are added in the submodules. How should I proceed further with this top module testing?

Look at `riscv-formal` run it understand how it works.  We can think about doing something similar.

> Also, what should I add to git actions ci?

Look at the file `.github/workflows/ci.yml` please add a step to install yosys symbiyosis and run the `make -C bench/formal` tests.

Next time when you comment.
Its better to respond to the comments where I made them, I cannot recall some of the context of these questions. Others in the future will have a really hard time to see the context.

rtl/verilog/mor1kx_cpu_cappuccino.v
```diff
@@ -1585,4 +1585,138 @@ module mor1kx_cpu_cappuccino
...
+         assert (carry_set_o !== 1'bx && carry_clear_o !== 1'bx);
+         assert (overflow_set_o !== 1'bx && overflow_clear_o !== 1'bx);
+      end
+
+   //SPR Properties check------------->
+   always @(posedge clk)
+      if (f_past_valid && !$past(rst) && ($rose(spr_bus_ack_dmmu_i)
+          || $rose(spr_bus_ack_immu_i) ||
+          $rose(spr_bus_ack_dc_i) || $rose(spr_bus_ack_ic_i)))
+         assert (spr_bus_stb_o || $past(spr_bus_stb_o));
```
stffrdhrn, line 1702: I think all of these SPR bus property checks, here and in other modules should go into a SPR interface module. 
See how this can be done here:
https://github.com/ZipCPU/zipcpu/blob/master/rtl/core/pfcache.v#L692
Example module:
https://github.com/ZipCPU/zipcpu/blob/master/rtl/ex/fwb_master.v

That is where we put assumes i.e.

```
		// Assume the slave will respond within F_MAX_ACK_DELAY cycles,
		// counted either from the end of the last request, or from the
		// last ACK received
```

rtl/verilog/mor1kx_icache.v
```diff
@@ -483,4 +483,196 @@ module mor1kx_icache
...
+   always @(posedge clk)
+      if (invalidate_o)
+         assert (spr_bus_addr_i == `OR1K_SPR_ICBIR_ADDR);
+
+   //SPR ack is seen only if strobe and write signal rise in the previous clock cycle
+   always @(posedge clk)
+      if ($rose(spr_bus_ack_o) && f_past_valid && !$past(rst))
+         assert ($past(spr_bus_stb_i) & $past(spr_bus_we_i)
+                 & $past(spr_bus_addr_i) == `OR1K_SPR_ICBIR_ADDR
+                 & !$past(refill_o));
```
stffrdhrn, line 649: I think its better to put these in the spr interface and assert the ack comes after the request signals.

stffrdhrn (commented):
We still need to think about using interface files to make some of these assumptions/asserts more module.   This short review provides some pointers.

rtl/verilog/mor1kx_icache.v
Harshitha172000: I can add this to spr interface. But in some modules spr ack won't arrive right after spr_stb. In dcache/icache, ack may arrive two-three clock cycles after spr strobe. In dmmu/immu, ack is always seen in the next clock after spr strobe.

rtl/verilog/mor1kx_cpu_cappuccino.v
Harshitha172000: Some spr properties are not common to all spr slaves. Dcache/Icache spr acknowledgements don't satisfy reset property but dmmu/immu does.
```
`define SLAVE_ASSUME assert
`define SLAVE_ASSERT assume

  reg f_past_valid;
  initial f_past_valid = 0;
  always @(posedge clk)
     f_past_valid <= 1'b1;
  always @(*)
     if (!f_past_valid)
        `SLAVE_ASSUME (rst);

  initial `SLAVE_ASSUME (!spr_bus_ack_o); //Fails in icache/dcache
``` 
Do we need to fix cache "not to acknowledge" during the reset? Or Is it ok to have spr slaves with different properties?

stffrdhrn (commented):
In your description you write.

```
Branch module: Remove f_past_valid from all branch modules and add f_not_unknown
               to assert branch properties.

Top module: ASSUME to ASSERT definations are changed to ASSUME only in few modules.
            Assertions based on input tends to fail during the top module test because
            external signals that drive inputs have no scope in the submodule. So, solver
            traces these external signals to fail assertion.
```

This description seems to cover changes that have been done since your first commit.  This kind of thing should be comments in the PR.  Your commit description should explain what the commit is any why.

The first sentence is good: `Added FORMAL directives to cappuccino modules plus sby files to run the test.`

Next you can explain some assumptions/or design choices. I.e.

This first patch adds assertions withouth impacting any core/module logic ...

All verification passes using xxx versions of xxx software.

For sub-module testing we use the approach of .... as described in ...

rtl/verilog/mor1kx_cpu_cappuccino.v
stffrdhrn: I think its ok modules have different properties.  The interface file should confirm common behavior like ack comes after 3 clocks.  When reading data comes after x clocks.  

I think the initial state checks like state at reset is not so important.

rtl/verilog/mor1kx_icache.v
stffrdhrn: As we discussed, like the fwb_master interface we should not assert exactly when the assert comes. We should assert that it comes within x cycles.  That can be configured per instantiation

rtl/verilog/mor1kx_decode.v
```diff
@@ -537,4 +537,169 @@ module mor1kx_decode
...
+                                  decode_op_branch_o};
+
+  //one instruction => one opcode
+  always @(*)
+     assert ($onehot0(f_set_of_opcodes));
+
+  //Load opcodes cannot decode store instructions
+  always @(*)
+     if (decode_op_lsu_load_o)
+        assert (!decode_op_lsu_store_o);
```
stffrdhrn, line 568: Is this needed? You already cover this in the onehot assert above.

Harshitha172000: No, I will remove that part.

Harshitha172000:
Edited PR with the following changes:

1.  Removed reset assertions on internal state variables.
2.  Moved assertions from the CPU module to respective submodules.
3.  Reduced internal state assertions up to some extent.

Submodules still have spr assertions. I will clean it up once I add the spr interface.

rtl/verilog/mor1kx_simple_dpram_sclk.v
```diff
@@ -69,4 +69,49 @@ endgenerate
...
+   //Output data should remain unchanged if there is no read enable
+   always @(posedge clk)
+      if (f_past_valid && !$past(re))
+         assert ($stable(dout));
+
+   //Output data changes only if there is read or write enable
+   always @(posedge clk)
+      if (f_past_valid && $changed(dout))
+         assert ($past(re) | $past(we));
+`endif
```
stffrdhrn, line 115: This looks good.

rtl/verilog/mor1kx_true_dpram_sclk.v
```diff
@@ -57,4 +57,83 @@ module mor1kx_true_dpram_sclk
...
+   always @(posedge global_clock)
+      if (f_past_valid && $rose(clk_a) && $fell(clk_b) && $past(we_a))
+         assert ($stable(dout_b));
+
+   //Port A shouldn't be affected by write signals of port B
+   always @(posedge global_clock)
+      if (f_past_valid && $rose(clk_b) && $fell(clk_a) && $past(we_b))
+         assert ($stable(dout_a));
+
+`endif
```
stffrdhrn, line 138: This looks good

rtl/verilog/mor1kx_cache_lru.v
```diff
@@ -277,5 +277,19 @@ module mor1kx_cache_lru(/*AUTOARG*/
...
+   always @(*) begin
+      assert (NUMWAYS > 1);
+`ifdef LRU
+      assume ($onehot0(access));
+`endif
+      assert ($onehot(lru_pre));
+      assert ($onehot(lru_post));
+   end
+
+`endif
```
stffrdhrn, line 293: This looks good.

rtl/verilog/mor1kx_decode.v
```diff
@@ -537,4 +537,164 @@ module mor1kx_decode
...
+                 decode_opc_insn_o == `OR1K_OPCODE_LHZ ||
+                 decode_opc_insn_o == `OR1K_OPCODE_LHS);
+      else if (decode_lsu_length_o == 2 && (decode_op_lsu_store_o
+               || decode_op_lsu_load_o))
+         assert (decode_opc_insn_o == `OR1K_OPCODE_SW ||
+                 decode_opc_insn_o == `OR1K_OPCODE_SWA ||
+                 decode_opc_insn_o == `OR1K_OPCODE_LWS ||
+                 decode_opc_insn_o == `OR1K_OPCODE_LWZ ||
+                 decode_opc_insn_o == `OR1K_OPCODE_LWA ||
+                 decode_opc_insn_o == `OR1K_OPCODE_LD);
```
stffrdhrn, line 595: This looks ok, but should this be the other way around?

```
         if (decode_opc_insn_o == `OR1K_OPCODE_SB ||
                 decode_opc_insn_o == `OR1K_OPCODE_LBZ ||
                 decode_opc_insn_o == `OR1K_OPCODE_LBS)
            assert (decode_lsu_length_o == 0);
      else if (decode_opc_insn_o == `OR1K_OPCODE_SH ||
                 decode_opc_insn_o == `OR1K_OPCODE_LHZ ||
                 decode_opc_insn_o == `OR1K_OPCODE_LHS);
           assert (decode_lsu_length_o == 1);
      else
           assert (decode_lsu_length_o == 2);
```

Maybe not? if we do my way its just copying the assign statements.

rtl/verilog/mor1kx_decode.v
```diff
@@ -537,4 +537,164 @@ module mor1kx_decode
...
+   end
+
+   //Only Branch Instructions like l.jal, l.jr, l.jalr, l.j,
+   //can set decode_op_branch_o high.
+   always @(*) begin
+      if (decode_op_branch_o)
+         assert (decode_opc_insn_o < `OR1K_OPCODE_NOP ||
```
stffrdhrn, line 602: That is this `nop` for?

rtl/verilog/mor1kx_decode.v
```diff
@@ -537,4 +537,164 @@ module mor1kx_decode
...
 
+
+
+/*----------------------Formal Checking---------------------*/
+
+`ifdef FORMAL
+
+    // Opcode ==> decode_insn_i[31:26]
+    always @(*)
+       assert (decode_opc_insn_o == decode_insn_i[`OR1K_OPCODE_SELECT]);
```
stffrdhrn, line 548: This is just verifying an assign, maybe not so helpful.

rtl/verilog/mor1kx_decode.v
```diff
@@ -537,4 +537,164 @@ module mor1kx_decode
...
+   // If instruction fetched is of double type and operand
+   // width is not of 64 bits, then its illegal
+   always @(*)
+      if (OPTION_OPERAND_WIDTH != 64 && decode_opc_insn_o
+          == `OR1K_OPCODE_LD || decode_opc_insn_o == `OR1K_OPCODE_SD)
+         assert (decode_except_illegal_o);
+
+   //Valid instructions shouldn't result in illegal exception
+   always @(*)
+      if (decode_opc_insn_o == f_set_of_opcodes)
```
stffrdhrn, line 625: Comparing `decode_opc_insn_o` with `f_set_of_opcodes` doesn't make sense to me.  Can you explain?

rtl/verilog/mor1kx_decode.v
```diff
@@ -537,4 +537,164 @@ module mor1kx_decode
...
+         assert (!decode_except_illegal_o);
+
+   always @(*) begin
+      //Illegal to decode FPU instruction when FPU feature is disabled.
+      if ((FEATURE_FPU == "NONE" | decode_insn_i[`OR1K_FPUOP_DOUBLE_BIT]) &&
```
stffrdhrn, line 630: Why do we need the `double` condition check?  This says to me , if FPU is disabled or insn is a double insn, it seems unrelated.

rtl/verilog/mor1kx_decode.v
```diff
@@ -537,4 +537,164 @@ module mor1kx_decode
...
+	       FEATURE_TRAP=="ENABLED") ||
+	      (decode_insn_i[`OR1K_SYSTRAPSYNC_OPC_SELECT] ==
+	       `OR1K_SYSTRAPSYNC_OPC_MSYNC) ||
+	      (decode_insn_i[`OR1K_SYSTRAPSYNC_OPC_SELECT] ==
+	       `OR1K_SYSTRAPSYNC_OPC_PSYNC &&
+	       FEATURE_PSYNC!="NONE") ||
+	      (decode_insn_i[`OR1K_SYSTRAPSYNC_OPC_SELECT] ==
+	       `OR1K_SYSTRAPSYNC_OPC_CSYNC &&
+	       FEATURE_CSYNC!="NONE")))
+         assert (!decode_except_illegal_o);
```
stffrdhrn, line 683: In general I think all of the `decode_except_illegal_o` may be a bit too much as its just combinational logic verification, its kind of just a copy of above.

rtl/verilog/mor1kx_decode.v
```diff
@@ -537,4 +537,164 @@ module mor1kx_decode
...
+
+   //Decoded subtraction instruction should drive decode_adder_do_sub_o
+   always @(*)
+      if (decode_adder_do_sub_o)
+         assert (decode_insn_i[`OR1K_ALU_OPC_SELECT] == `OR1K_ALU_OPC_SUB &
+                 decode_insn_i[`OR1K_OPCODE_SELECT] == `OR1K_OPCODE_ALU
+                 || decode_op_setflag_o);
```
stffrdhrn, line 690: This looks like a copy of this.  Maybe not so helpful.  Also why do you use `||` vs `|` logical vs bitwise.

```
   assign decode_adder_do_sub_o = (opc_insn == `OR1K_OPCODE_ALU &
				   opc_alu == `OR1K_ALU_OPC_SUB) |
				  decode_op_setflag_o;
```

rtl/verilog/mor1kx_decode.v
```diff
@@ -537,4 +537,164 @@ module mor1kx_decode
...
+
+   //ADDC instruction should set output decode_adder_do_carry_o high
+   always @(*)
+      if (decode_adder_do_carry_o)
+         assert (decode_opc_insn_o == `OR1K_OPCODE_ADDIC ||
+                (FEATURE_ADDC!="NONE" && ((opc_insn == `OR1K_OPCODE_ALU
+                & opc_alu ==`OR1K_ALU_OPC_ADDC) )));
+
```
stffrdhrn, line 698: Summary, I think mostly only the `onehot0` check is useful.  The others seem to just be copies of the assign statement.  What do you think?

Harshitha172000: I think the same, that's why I wanted to clean up `decode_except_illegal_o` assertions. Other than `onehot0` most of them reflect assign statements.

rtl/verilog/mor1kx_decode.v
Harshitha172000: Should I keep this assertion? Or clean it up?

rtl/verilog/mor1kx_branch_predictor_gshare.v
```diff
@@ -113,4 +113,26 @@ module mor1kx_branch_predictor_gshare
...
+                        flag_i !== 1'bx &&
+                        prev_op_brcond_i !== 1'bx &&
+                        branch_mispredict_i !== 1'bx &&
+                        ^brn_pc_i !== 1'bx);
+
+   initial f_not_unknown = 0;
+
+   always @(*)
+      if (f_not_unknown)
+         assert (predicted_flag_o !== 1'bx);
```
stffrdhrn, line 135: I think we should remove this unknown state validation.  I am not sure what it really helps.

However, I think we should try to add some kind of validation for the gshare FSM.  I cant really see what could be unconstrained though.

Perhaps we can completely remove this now.

rtl/verilog/mor1kx_branch_predictor_saturation_counter.v
```diff
@@ -94,4 +94,25 @@ module mor1kx_branch_predictor_saturation_counter
...
+                        flag_i !== 1'bx &&
+                        padv_decode_i !== 1'bx &&
+                        branch_mispredict_i !== 1'bx &&
+                        prev_op_brcond_i !== 1'bx);
+
+   initial f_not_unknown = 0;
+
+   always @(*)
+      if (f_not_unknown)
+         assert (predicted_flag_o !== 1'bx);
```
stffrdhrn, line 115: I think we should remove this unknown state validation. I am not sure what it really helps.

However, I think we should try to add some kind of validation for the saturation FSM. I cant really see what could be unconstrained though.

Perhaps we can completely remove this now.

stffrdhrn (commented):
Review of branch prediction modules.  I think its better to remove all the formal checks the do not seem to really do much and just add noise.

rtl/verilog/mor1kx_branch_prediction.v
```diff
@@ -110,4 +110,30 @@ end else begin
 end
 endgenerate
 
+/*------Formal Checking------*/
+
+`ifdef FORMAL
+
+   reg f_not_unknown = (op_bf_i !== 1'bx &&
```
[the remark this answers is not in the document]
Harshitha172000: Keeping it as a wire and initializing it to zero creates driver conflict. So, I have used reg.

rtl/verilog/mor1kx_branch_predictor_saturation_counter.v
Harshitha172000: I don't know what to validate in FSM as it has only internal state variables.

rtl/verilog/mor1kx_branch_prediction.v
```diff
@@ -110,4 +110,30 @@ end else begin
...
+
+   initial f_not_unknown = 0;
+
+   //Output shouldn't have unknown bits.
+   always @(*) begin
+      if (f_not_unknown) begin
+         assert (branch_mispredict_o !== 1'bx);
+         assert (predicted_flag_o !== 1'bx);
+      end
+   end
```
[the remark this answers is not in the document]
stffrdhrn: Actually, if we do need formal verification in branch prediction it should only be here at the top level.

rtl/verilog/mor1kx_bus_if_wb32.v
```diff
@@ -180,4 +180,80 @@ module mor1kx_bus_if_wb32
...
+
+   always @(*)
+      if (BUS_IF_TYPE == "B3_REGISTERED_FEEDBACK") begin
+         assert (wbm_adr_o == cpu_adr_i);
+         assert (wbm_stb_o == cpu_req_i);
+         assert (wbm_sel_o == cpu_bsel_i);
+         assert (wbm_dat_o == cpu_dat_i);
+         assert (cpu_err_o == wbm_err_i);
+         assert (cpu_ack_o == wbm_ack_i);
+      end
```
stffrdhrn, line 205: I dont see this check as very helpful.

rtl/verilog/mor1kx_bus_if_wb32.v
```diff
@@ -180,4 +180,80 @@ module mor1kx_bus_if_wb32
...
+    always @(posedge clk) begin
+       if (BUS_IF_TYPE != "B3_REGISTERED_FEEDBACK" && BUS_IF_TYPE != "B3_READ_BURSTING") begin
+          if (f_past_valid && $past(rst))
+             assert (!wbm_stb_o);
+          if (f_past_valid && !$past(rst) && ($past(cpu_ack_o) | $past(cpu_err_o)))
+              assert (!wbm_stb_o);
+          if (f_past_valid && !$past(rst) && (!$past(cpu_ack_o) & !$past(cpu_err_o)) && cpu_req_i)
+             assert (wbm_stb_o);
+       end
+    end
```
stffrdhrn, line 216: Can you add a comment about what you are trying to verify here?

rtl/verilog/mor1kx_bus_if_wb32.v
```diff
@@ -180,4 +180,80 @@ module mor1kx_bus_if_wb32
...
+         assert (^cpu_dat_o !== 1'bx);
+         assert (^wbm_adr_o !== 1'bx);
+         assert (^wbm_stb_o !== 1'bx);
+         assert (wbm_cyc_o !== 1'bx);
+         assert (^wbm_sel_o !== 1'bx);
+         assert (wbm_we_o !== 1'bx);
+         assert (^wbm_cti_o !== 1'bx);
+         assert (^wbm_bte_o !== 1'bx);
+         assert (^wbm_dat_o !== 1'bx);
+      end
```
stffrdhrn, line 256: The reset of these I dont see as very helpful.

stffrdhrn (commented):
Checked bus_if_wb32.

It looks like this is mostly just checking assigns.  In the module there are somethings we should try to guarantee through formal I think.

We should confirm that:
  - `wbm_cti_o` tag type should be only `010` incrementing or `111` end-of-burst . 
  - In burst mode we will also see `wbm_cti_o` as `000` classic
  - In burst mode if we see a `010` we should always see a `111` after `wbm_ack_i` (assume ack is held high until we see `111`).

rtl/verilog/mor1kx_branch_prediction.v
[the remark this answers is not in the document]
Harshitha172000: Here, there are only two output ports and I can't assert more properties. But if I verify from the CPU level, I can also assert properties based on its input interface.

rtl/verilog/mor1kx_cpu_cappuccino.v
```diff
@@ -1585,4 +1585,97 @@ module mor1kx_cpu_cappuccino
...
+
+   //Without spr strobe, spr_bus_dat_o remains unchanged.
+   always @(posedge clk) begin
+      if (f_past_valid && !$past(rst) && !$past(spr_bus_stb_o)
+          && !spr_bus_stb_o && !rst && !$past(spr_bus_stb_o,2)
+          && !$past(rst,2)) begin
```
stffrdhrn, line 1668: It may be a bit easier to read as:

```
     if (f_past_valid
        && !rst && !spr_bus_stb_o
        && !$past(rst) && !$past(spr_bus_stb_o)
        && !$past(rst,2) && !$past(spr_bus_stb_o,2)) begin
```

stffrdhrn (commented):
`mor1kx_cpu_cappuccino` look ok just a few comments.

stffrdhrn (commented):
You have added `fspr_master` and `fspr_slave` to this commit.  On the call I asked you to keep this on a separate branch i.e. `formal-spr` like you did with `formal-ci`.  That way we can review this separately.

Can you remove this from the commit and do the work on a separate PR?

rtl/verilog/mor1kx_bus_if_wb32.v
Harshitha172000: Will be removed.

rtl/verilog/mor1kx_bus_if_wb32.v
Harshitha172000: Ok, I will remove it. 
I will check the properties mentioned above.
I think I should move all wishbone properties to wb interface.

Harshitha172000:
> You have added `fspr_master` and `fspr_slave` to this commit. On the call I asked you to keep this on a separate branch i.e. `formal-spr` like you did with `formal-ci`. That way we can review this separately.
> 
> Can you remove this from the commit and do the work on a separate branch/PR?

I have removed my last commit and moved the spr interface to `formal-spr` PR.

rtl/verilog/mor1kx_cpu_cappuccino.v
```diff
@@ -1585,4 +1585,72 @@ module mor1kx_cpu_cappuccino
...
+         assert ($stable(spr_bus_dat_dmmu_i));
+         assert ($stable(spr_bus_dat_immu_i));
```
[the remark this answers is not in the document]
Harshitha172000: Nothing depends on it. I added this to verify that the spr read data port is not affected by any other signals other than strobe.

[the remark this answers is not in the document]
stffrdhrn: I am not sure this is needed then, as without strobe I think nothing should care about the `spr_bus_dat_*mmu_i` signals.

rtl/verilog/mor1kx_cpu_cappuccino.v
```diff
@@ -1585,4 +1585,72 @@ module mor1kx_cpu_cappuccino
...
+      end
+   end
+
+   always @(posedge clk)
+      if (f_past_valid && !$past(rst) && !rst && $stable(spr_bus_addr_o)
+          && $past(spr_bus_stb_o) && !$past(spr_bus_stb_o,2) && $past(rst,2))
```
stffrdhrn, line 1651: Can you add a comment on this?  Why do you have this case only after reset? `$past(rst,2)`

rtl/verilog/mor1kx_decode.v
```diff
@@ -537,4 +537,26 @@ module mor1kx_decode
...
+                                   decode_op_div_signed_o,
+                                   decode_op_ext_o, decode_op_div_unsigned_o,
+                                   decode_op_shift_o, decode_op_ffl1_o,
+                                   decode_op_movhi_o, decode_op_msync_o,
+                                   decode_op_branch_o};
+
+   //one instruction => one opcode
+   always @(*)
+      assert ($onehot0(f_set_of_opcodes));
+
```
stffrdhrn, line 560: OK.

rtl/verilog/mor1kx_cpu_cappuccino.v
Harshitha172000: It's typo mistake, it should be `!$past(rst,2 ) && $past(spr_bus_stb_o,2)`. Anyhow these spr assertions will be reformed in spr interface.

rtl/verilog/mor1kx_decode.v
Harshitha172000: I think wire `f_set_of_opcodes` should be used instead of reg.

rtl/verilog/mor1kx_cpu_cappuccino.v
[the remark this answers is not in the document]
Harshitha172000: Ok, I will remove this.

rtl/verilog/mor1kx_cpu_cappuccino.v
Harshitha172000: This `onehot` assertion fails.

image

rtl/verilog/mor1kx_decode.v
stffrdhrn: ok.

rtl/verilog/mor1kx_execute_alu.v
```diff
@@ -814,7 +814,7 @@ endgenerate
 			 op_movhi_i ? immediate_i :
 			 op_ext_i ? ext_result :
 			 op_mul_i ? mul_result[OPTION_OPERAND_WIDTH-1:0] :
-			 fpu_arith_valid ? fpu_result :
+			 fpu_arith_valid ? fpu_result :
```
stffrdhrn, line 817: Can you revert the whitespace changes in this file?  Making changes like this messes up the blame history.

rtl/verilog/mor1kx_wb_mux_cappuccino.v
```diff
@@ -52,4 +52,47 @@ module mor1kx_wb_mux_cappuccino
...
+   reg f_not_unknown = (^alu_result_i !== 1'bx && ^lsu_result_i !== 1'bx &&
+                        ^mul_result_i !== 1'bx && ^spr_i !== 1'bx &&
+                        op_mul_i !== 1'bx && op_lsu_load_i !== 1'bx &&
+                        op_mfspr_i !== 1'bx);
+   initial f_not_unknown = 1'b0;
+
+   always @*
+      if (f_not_unknown)
+         assert (^rf_result_o !== 1'bx);
+
```
stffrdhrn, line 74: I am not sure we need this `f_not_unknown` checks.

rtl/verilog/mor1kx_wb_mux_cappuccino.v
```diff
@@ -52,4 +52,47 @@ module mor1kx_wb_mux_cappuccino
...
+   reg f_past_valid;
+   initial f_past_valid = 0;
+   always @(posedge clk)
+      f_past_valid <= 1'b1;
+   always @(*)
+      if (!f_past_valid)
+         assume (rst);
+   always @(posedge clk)
+      if (f_past_valid && !$past(rst))
+         `ASSUME ($onehot0({op_mul_i, op_lsu_load_i, op_mfspr_i}));
```
stffrdhrn, line 84: This looks ok

rtl/verilog/mor1kx_wb_mux_cappuccino.v
```diff
@@ -52,4 +52,47 @@ module mor1kx_wb_mux_cappuccino
...
+
+   always @(posedge clk) begin
+      if (f_past_valid && !$past(rst) && !$past(op_mul_i)) begin
+         if ($past(op_mfspr_i))
+            assert (rf_result_o == $past(spr_i));
+         else if ($past(op_lsu_load_i))
+            assert (rf_result_o == $past(lsu_result_i));
+         else
+            assert (rf_result_o == $past(alu_result_i));
+      end
```
stffrdhrn, line 94: This just looks like a copy of above. 

The logic is:
   - op_mul_i  - forward next mul_result_i  next cycle
   - op_mfspr_i - forward this spr_i next cycle
   - op_lsu_load_i - forward this lsu_result_i next cycle
   - alu_result_i - forward this alu_result_i next cycle

So mul_result_i will come late, I think you can write it like:

```
      if (f_past_valid && !$past(rst)) begin
          if ($past(op_mfspr_i))
             assert (rf_result_o == $past(spr_i));
          else if ($past(op_lsu_load_i))
             assert (rf_result_o == $past(lsu_result_i));
          else if ($past(op_mul_i))
             assert (rf_result_o == op_mul_i);
          else
             assert (rf_result_o == $past(alu_result_i));
       end
```

rtl/verilog/mor1kx_ticktimer.v
```diff
@@ -87,4 +87,43 @@ module mor1kx_ticktimer
...
+   initial f_past_valid = 0;
+   initial assume(rst);
+   always @(posedge clk)
+      f_past_valid <= 1;
+   always @(*)
+      if (!f_past_valid)
+         assume (rst);
+
+   always @*
+      `ASSUME ($onehot0(spr_access_i));
```
stffrdhrn, line 110: what is this for?

rtl/verilog/mor1kx_pic.v
```diff
@@ -139,4 +139,45 @@ module mor1kx_pic
...
+   initial f_past_valid = 0;
+   initial assume(rst);
+   always @(posedge clk)
+      f_past_valid <= 1;
+   always @(*)
+      if (!f_past_valid)
+         assume (rst);
+
+   always @*
+      `ASSUME ($onehot0(spr_access_i));
```
stffrdhrn, line 162: what is this for?

rtl/verilog/mor1kx_ticktimer.v
Harshitha172000: This confirms one spr access at a time. As it's input here, so used assume to assert.

rtl/verilog/mor1kx_wb_mux_cappuccino.v
Harshitha172000: > This just looks like a copy of above.

I know, but I had no idea whatelse could be asserted.

rtl/verilog/mor1kx_wb_mux_cappuccino.v
Harshitha172000: As it just adds noise, I can remove it.

Harshitha172000:
I have added unknown tests in ALU and PCU too. Shall I remove those tests?

rtl/verilog/mor1kx_ticktimer.v
stffrdhrn: but there is only one input; one wire.  Its just saying spr_access_i should be 1 or 0, which is kind of given.

rtl/verilog/mor1kx_wb_mux_cappuccino.v
stffrdhrn: I think we should have a rule of "when in doubt, leave it out".  But did you get my point about adding `op_mul_i` to the assertion if/else?

rtl/verilog/mor1kx_cpu_cappuccino.v
stffrdhrn: Re `onehot` failing, ok we can look at that with the spr-ibterface work.

stffrdhrn:
> I have added unknown tests in ALU and PCU too. Shall I remove those tests?

I think so, I would like them to be removed everywhere for now.

rtl/verilog/mor1kx_bus_if_wb32.v
```diff
@@ -180,4 +180,56 @@ module mor1kx_bus_if_wb32
...
+   // 111- end of burst
+   // 010- incrementing
+   // 000- classic
+   always @*
+      assert (wbm_cti_o == 3'b111 | wbm_cti_o == 3'b010 | wbm_cti_o == 3'b000);
+
+   //If cycle type identifier is incrementing and wb ack arrives, then burst should be ended.
+   always @(posedge clk)
+      if (f_past_valid && !$past(rst) && $past(wbm_ack_i) && $past(wbm_cti_o) == 2'b010)
+         assert (wbm_cti_o == 3'b111);
```
stffrdhrn, line 233: ok

rtl/verilog/mor1kx_wb_mux_cappuccino.v
stffrdhrn: > I think we should have a rule of "when in doubt, leave it out". But did you get my point about adding `op_mul_i` to the assertion if/else?

OK, I see you added it,

rtl/verilog/mor1kx_wb_mux_cappuccino.v
```diff
@@ -52,4 +52,38 @@ module mor1kx_wb_mux_cappuccino
...
+         else if ($past(op_lsu_load_i))
+            assert (rf_result_o == $past(lsu_result_i));
+         else if ($past(op_mul_i))
+            assert (rf_result_o == mul_result_i);
+         else
+            assert (rf_result_o == $past(alu_result_i));
+      end
+   end
+
+`endif
```
stffrdhrn, line 88: ok.

rtl/verilog/mor1kx_pcu.v
```diff
@@ -137,4 +137,37 @@ module mor1kx_pcu
...
+   //PCMR and PCCR access requests are mutually exclusive
+   always @(*)
+      if (pcu_pccr_access)
+         assert (!pcu_pcmr_access);
+
+   reg f_not_unknown = spr_access_i !== 1'bx;
+   initial f_not_unknown = 0;
+   always @*
+      if (f_not_unknown)
+         assert (^spr_dat_o !== 1'bx);
```
stffrdhrn, line 171: Maybe we could confirm some thing with `pcu_pccr`
 i.e. if we have any inputs high, then `pcu_pccr` should increment.  It we wrap around it goes back to zero.

 If all inputs are low  `pcu_pccr` stays stable.

stffrdhrn (commented):
Comments on `ticktimer`, `pic` and `pcu`. These are are similar SPR functional models.

rtl/verilog/mor1kx_rf_cappuccino.v
```diff
@@ -362,4 +362,35 @@ end else begin
...
+      if (!f_past_valid)
+         assume (rst);
+
+   reg f_not_unknown = (^ctrl_alu_result_i !== 1'bx);
+   initial f_not_unknown = 0;
+   always @(*)
+      if (f_not_unknown) begin
+         assert (^decode_rfa_o !== 1'bx && ^decode_rfb_o !== 1'bx);
+         assert (^execute_rfa_o !== 1'bx && ^execute_rfb_o !== 1'bx );
+      end
```
stffrdhrn, line 384: Can you remove this check?

rtl/verilog/mor1kx_store_buffer.v
```diff
@@ -86,4 +86,95 @@ module mor1kx_store_buffer
...
+   always @(posedge clk)
+      if ($past(rst) && f_past_valid)
+         assert ((write_pointer == 0) && (read_pointer == 0)
+                 && (empty_o == 1) && (full_o == 0) &&
+                 (f_curr_fifo_entries == 0));
+
+   //Full FIFO Assertions------------>
+   //1. When FIFO is full, full flag should be set.
+   always @(posedge clk `OR_ASYNC_RST)
+      if (f_past_valid && (f_curr_fifo_entries > f_total_fifo_entries-1))
```
stffrdhrn, line 120: Why do you use a '>' comparison here.  Can you just say `f_curr_fifo_entries == f_total_fifo_entries`

rtl/verilog/mor1kx_store_buffer.v
```diff
@@ -86,4 +86,95 @@ module mor1kx_store_buffer
...
+   // write_i occurs, then full_o has to be set.
+   always @(posedge clk)
+      if (f_past_valid && ($past(f_curr_fifo_entries) ==
+          f_total_fifo_entries - 1) && $past(write_i) && !$past(read_i))
+         assert (full_o);
+
+   //4. When FIFO is full, no write_i.
+   always @(posedge clk) begin
+      if (f_past_valid && full_o)
+         assert (!write_i);
```
stffrdhrn, line 138: should this be an assumption?

rtl/verilog/mor1kx_store_buffer.v
```diff
@@ -86,4 +86,95 @@ module mor1kx_store_buffer
...
+   //Empty FIFO Assertions---------->
+   //1. When FIFO has no entries, empty flag has to be set.
+   always @(posedge clk)
+      if (f_curr_fifo_entries == 0 && f_past_valid)
+         assert (empty_o);
+
+   //2. If empty flag is set, full flag should be zero.
+   always @(posedge clk)
+      if (empty_o && f_past_valid)
+         assert (!full_o);
```
stffrdhrn, line 150: Can we just use `$onehot0` for empty_o / full_o?

rtl/verilog/mor1kx_store_buffer.v
```diff
@@ -86,4 +86,95 @@ module mor1kx_store_buffer
...
+   //3. When FIFO has one entry, and read occurs, then empty flag should be set.
+   always @(posedge clk)
+      if ($past(f_curr_fifo_entries == 1) && f_past_valid
+          && $past(read_i) && !$past(write_i))
+         assert (empty_o);
+
+   //4. When FIFO is empty, no read_i.
+   always @(posedge clk)
+      if (empty_o && f_past_valid && !write_i)
+         assert (!read_i);
```
stffrdhrn, line 161: Should this be an assume? Or is it ok to assert inputs like this?

rtl/verilog/mor1kx_store_buffer.v
```diff
@@ -86,4 +86,95 @@ module mor1kx_store_buffer
...
+   always @(posedge clk)
+      if (f_past_valid && !$past(write_i))
+         assert ($stable(write_pointer));
+
+   //Read pointer should not change when read_i is 0.
+   always @(posedge clk)
+      if (f_past_valid && !$past(read_i))
+         assert ($stable(read_pointer));
+
+`endif
```
stffrdhrn, line 179: OK, this fifo verification is nice, did you get any points from anywhere on how to do the verification?

rtl/verilog/mor1kx_store_buffer.v
Harshitha172000: ok

rtl/verilog/mor1kx_store_buffer.v
Harshitha172000: There is a major mistake in this test, I haven't set `f_past_valid` which's why this test is passing. I will fix it.

rtl/verilog/mor1kx_store_buffer.v
Harshitha172000: Yeah

rtl/verilog/mor1kx_pcu.v
Harshitha172000: `pcu_pccr` has a set of registers and genvar `pcu_num` is used to update particular reg in `pcu_pccr` based on inputs. It's difficult to check if a particular `pcu_pccr` register is incremented as there is no scope of `pcu_num` outside generate block and the condition to update events in `pcu_pccr` also depends on `pcu_num`.

Harshitha172000: > If all inputs are low `pcu_pccr` stays stable.

I tried to check this and I see the following trace.

image

I have added `!$past(spr_we_i) && !$past(spr_sys_mode_i)` in the `if` condition but solver fails this assert. `pcu_pccr` changes
due to spr write and system mode signals.
The solver is not adding the required constraint as specified in the `if` condition. I am not sure why this is happening.

stffrdhrn: Can you put this check on a branch in a separate commit?  I can have a look.  
In terms of `pcu_num` I think we can use use an `(* anyconst *)` and assume to assume the the register we are looking at.  There are 8 registers that  are all the same.

rtl/verilog/mor1kx_execute_alu.v
```diff
@@ -850,5 +850,192 @@ endgenerate
...
+   end
+
+   always @(posedge clk) begin
+      if (!$past(op_mul_i,4) && $past(op_mul_i,3) &&
+          f_past_valid && f_not_unknown
+          && FEATURE_MULTIPLIER == "THREESTAGE" && !$past(rst) 
+          && !$past(rst,2) && !$past(rst,3) && mul_valid &&
+          $past(decode_valid_i,4) && !$past(rst,4)) begin
+         assert ($past(alu_stall));
+         assert (!mul_unsigned_overflow);
```
stffrdhrn, line 899: why do we assert no overflow?  We should assert mul_valid here.  And also alu_stall is low.

rtl/verilog/mor1kx_execute_alu.v
```diff
@@ -850,5 +850,192 @@ endgenerate
...
+   end
+`endif
+
+   //Pipelined Multiplication
+   always @(posedge clk) begin
+      if ($past(decode_op_mul_i,3) && $past(padv_decode_i,3) &&
+          $past(padv_execute_i,2) && f_past_valid &&
+          FEATURE_MULTIPLIER == "PIPELINED" && !$past(rst)
+          && op_mul_i && !$past(rst,2) && !$past(rst,3)) begin
+         assert (!mul_unsigned_overflow);
```
stffrdhrn, line 910: Why assert no overflow?

rtl/verilog/mor1kx_execute_alu.v
```diff
@@ -850,5 +850,192 @@ endgenerate
...
+         assert (mul_valid);
+      end
+   end
+
+   //Simulation based multiplier
+   always @(*) begin
+      if (FEATURE_MULTIPLIER == "SIMULATION" &&
+          FEATURE_DIVIDER == "NONE" && FEATURE_FFL1 == "NONE"
+          && FEATURE_FPU == "NONE") begin
```
stffrdhrn, line 919: why are you also checking DIVIDER/FPU and FFL1 in this multiply related check? Can you just check `op_mul_i`?

rtl/verilog/mor1kx_execute_alu.v
```diff
@@ -850,5 +850,192 @@ endgenerate
...
+         //No stall for simulation based multiplier
+         assert (!alu_stall);
+         assert (mul_valid);
+      end
+   end
+
+   always @(*) begin
+      if (FEATURE_MULTIPLIER == "NONE") begin
+         assert (!mul_result);
```
stffrdhrn, line 928: I think in this case we can do `assume(! op_mul_i)`

Harshitha172000:
Changes done:

1. Fixed issues with Store Buffer.
2. Added assertions to tick timer.
3. Removed unknown test from the register file, pcu, pic, and alu.

rtl/verilog/mor1kx_execute_alu.v
```diff
@@ -850,5 +850,173 @@ endgenerate
...
+      if (FEATURE_DIVIDER == "SIMULATION" &&
+         (opc_alu_i == `OR1K_ALU_OPC_DIV ||
+         opc_alu_i == `OR1K_ALU_OPC_DIVU)) begin
+         assert (div_valid);
+         //Divide by zero testing
+         if (b==0)
+            assert (div_by_zero);
+       end
+
+       if (FEATURE_DIVIDER == "NONE") begin
```
stffrdhrn, line 944: Here you can `assume(! op_div_i)`

rtl/verilog/mor1kx_execute_alu.v
```diff
@@ -850,5 +850,173 @@ endgenerate
...
+
+   always @(*) begin
+      if (FEATURE_MULTIPLIER == "NONE") begin
+         assert (!mul_result);
+         assert (!mul_unsigned_overflow);
+      end
+   end
+
+   //SIMULATION DIVIDER
+   always @(*) begin
```
stffrdhrn, line 934: We are missing the check for `FEATURE_DIVIDER=="SERIAL"` here.  For that case we could confirm that `alu_valid_o` comes after `OPTION_OPERAND_WIDTH` (32) clocks.

Note , we can also assume that all inputs are stable as long as `alu_stall` is high

rtl/verilog/mor1kx_execute_alu.v
```diff
@@ -850,5 +850,173 @@ endgenerate
...
+   reg [OPTION_OPERAND_WIDTH-1:0] f_sll = a << f_b;
+   reg [OPTION_OPERAND_WIDTH-1:0] f_srl = a >> f_b;
+
+   always @(*) begin
+      if (OPTION_SHIFTER == "BARREL") begin
+         //Logical shifts
+         if (opc_alu_shr == `OR1K_ALU_OPC_SECONDARY_SHRT_SLL && b == f_b)
+            assert (shift_result == f_sll);
+         if (opc_alu_shr == `OR1K_ALU_OPC_SECONDARY_SHRT_SRL && b == f_b)
+            assert (shift_result == f_srl);
```
stffrdhrn, line 961: Instead of validating `shift_result` could you verify `alu_result_o`?

rtl/verilog/mor1kx_execute_alu.v
```diff
@@ -850,5 +850,173 @@ endgenerate
...
+   end
+
+//-----------------ALU PROPERTIES------------------
+
+   always @(*) begin
+
+      if (!op_logic && !op_cmov && !op_movhi_i &&
+          !op_ext_i && !op_mul_i && !fpu_arith_valid
+          && !op_shift_i && !op_div_i && !op_ffl1_i)
+         assert (alu_result_o == adder_result_o);
```
stffrdhrn, line 972: what are you verifying here?  That the default is adder? I don't think its needed.

rtl/verilog/mor1kx_execute_alu.v
```diff
@@ -850,5 +850,173 @@ endgenerate
...
+
+      //If carry is generated then carry flag has to be set
+      if (adder_carryout & op_add_i)
+         assert (carry_set_o);
+
+      //Overflow set conditions
+      if (op_add_i && adder_signed_overflow && FEATURE_OVERFLOW != "NONE")
+         assert (overflow_set_o);
+      if (FEATURE_OVERFLOW != "NONE" && adder_signed_overflow && op_add_i)
+         assert (overflow_set_o);
```
stffrdhrn, line 986: are these two checks the same?

rtl/verilog/mor1kx_execute_alu.v
```diff
@@ -850,5 +850,173 @@ endgenerate
...
+
+      //Stall is observed while executing insn like mul, div, fpu, shift, ffl1.
+      if (!op_div_i && !op_mul_i && !fpu_op_is_arith &&
+          !fpu_op_is_cmp && !op_shift_i && !op_ffl1_i)
+         assert (!alu_stall);
+
+      //No stall when alu valid is set.
+      if (alu_valid_o)
+         assert (!alu_stall);
```
stffrdhrn, line 995: This is how its defined via assign, maybe this is not needed.

rtl/verilog/mor1kx_execute_alu.v
```diff
@@ -850,5 +850,173 @@ endgenerate
...
+      //Overflow set conditions
+      if (op_add_i && adder_signed_overflow && FEATURE_OVERFLOW != "NONE")
+         assert (overflow_set_o);
+      if (FEATURE_OVERFLOW != "NONE" && adder_signed_overflow && op_add_i)
+         assert (overflow_set_o);
+
+      //Stall is observed while executing insn like mul, div, fpu, shift, ffl1.
+      if (!op_div_i && !op_mul_i && !fpu_op_is_arith &&
+          !fpu_op_is_cmp && !op_shift_i && !op_ffl1_i)
+         assert (!alu_stall);
```
stffrdhrn, line 991: This comment seems a bit wrong, I think it should say something like if we dont have long operations we wont stall.

rtl/verilog/mor1kx_execute_alu.v
```diff
@@ -850,5 +850,173 @@ endgenerate
...
+   end
+
+   reg [OPTION_OPERAND_WIDTH-1:0] f_logic_and = a & b;
+   reg [OPTION_OPERAND_WIDTH-1:0] f_logic_or = a | b;
+   reg [OPTION_OPERAND_WIDTH-1:0] f_logic_xor = a ^ b;
+
+   always @(*) begin
+      //No stall when alu result is valid
+      if (alu_valid_o)
+         assert (!alu_stall);
```
stffrdhrn, line 1006: This is a repeat of the assert a few lines above, we can remove it.

stffrdhrn (commented):
Reviewed `ALU` quite a few comments. If you want to leave for later please just add TODO items and we can sort them out later.

rtl/verilog/mor1kx_pcu.v
Harshitha172000: `(* anyconst *) f_pcu_num` may not be same as `pcu_num`. The input event to be updated is dependent on `pcu_num`. The solver may assume some `f_pcu_num` but the actual updated `pcu_num` may be different. There are 11 events that will be updated based on `pcu_num`.

rtl/verilog/mor1kx_execute_alu.v
Harshitha172000: Why do we need this assume? We are not asserting anything which is related to this assumption.
You mean assume to assert ` ASSUME(!op_mul_i)` ?

rtl/verilog/mor1kx_execute_alu.v
Harshitha172000: Few implementations always set overflow 0, so I asserted it.

rtl/verilog/mor1kx_execute_alu.v
Harshitha172000: Few implementations always set overflow to 0. If it's not required, maybe I can remove it.

rtl/verilog/mor1kx_execute_alu.v
Harshitha172000: I tried to check this and stuck with an induction fail. After fixing it, I will add checks for the serial multiplier, divider, and shifter.

rtl/verilog/mor1kx_execute_alu.v
stffrdhrn: ok, that seems like just a missing functionality, can you remove that assert, its a bit misleading.

rtl/verilog/mor1kx_execute_alu.v
stffrdhrn: Yes. `ASSUME`

rtl/verilog/mor1kx_execute_alu.v
stffrdhrn: Can you add that as a TODO comment?

rtl/verilog/mor1kx_store_buffer.v
```diff
@@ -86,4 +86,92 @@ module mor1kx_store_buffer
...
+   always @(posedge clk)
+      f_past_valid <= 1;
+   always @(posedge clk)
+      if (!f_past_valid)
+         assume (rst);
+
+   localparam f_total_fifo_entries = 2 ** DEPTH_WIDTH;
+   wire [DEPTH_WIDTH:0]f_curr_fifo_entries;
+
+   assign f_curr_fifo_entries = write_pointer - read_pointer;
```
stffrdhrn, line 110: Previously you had an independent counter to track fifo entries, why did you remove that?  I think the previous counter was better as it allows for parallel verification of the pointer logic.

Harshitha172000: It was not tracking the number of FIFO entries correctly, and few assertions were failing so I used this.

rtl/verilog/mor1kx_execute_alu.v
Harshitha172000: Added

rtl/verilog/mor1kx_execute_alu.v
Harshitha172000: ok

rtl/verilog/mor1kx_store_buffer.v
stffrdhrn: ok

rtl/verilog/mor1kx_execute_ctrl_cappuccino.v
```diff
@@ -384,4 +384,133 @@ endgenerate
...
+          | $past(ctrl_except_dpagefault_o) && !$past(rst))
+         assert (!ctrl_op_lsu_load_o && !ctrl_op_lsu_store_o &&
+                 !ctrl_op_lsu_atomic_o);
+
+    always @(*) begin
+       if (execute_valid_o)
+          assert (!execute_stall);
+       if (ctrl_valid_o)
+          assert (!ctrl_stall);
+    end
```
stffrdhrn, line 494: I dont think we need these combinational aserts

rtl/verilog/mor1kx_execute_ctrl_cappuccino.v
```diff
@@ -384,4 +384,133 @@ endgenerate
...
+          assert ($stable(ctrl_op_mfspr_o));
+          assert ($stable(ctrl_op_mtspr_o));
+          assert ($stable(ctrl_op_rfe_o));
+          assert ($stable(ctrl_op_msync_o));
+          assert ($stable(ctrl_lsu_length_o));
+          assert ($stable(ctrl_lsu_zext_o));
+       end
+    end
+
+`endif
```
stffrdhrn, line 514: `exec_ctrl` looks ok, thanks.

rtl/verilog/mor1kx_execute_alu.v
Harshitha172000: This validation `ASSUME(!op_mul_i)` won't pass in the top module test, BMC fails. Even if `FEATURE_MULTIPLIER == "NONE"`, its possible to have instruction decoded is of division type.  Maybe we should not assert this.

rtl/verilog/mor1kx_execute_alu.v
Harshitha172000: Removed overflow and asserted alu_stall is low. Not asserting mul_valid because I have used that in the condition.

rtl/verilog/mor1kx_execute_alu.v
Harshitha172000: Done

rtl/verilog/mor1kx_execute_alu.v
Harshitha172000: Comment changed

Harshitha172000:
Added checks for the serial features like multiplier, divider, and shifter.  Let me know if the approach used is correct.
We need to update the SBY file when the feature type is changed. Shall I add tasks for all features? This may create too many `execute_alu` directories after the test.

rtl/verilog/mor1kx_execute_ctrl_cappuccino.v
Harshitha172000: Removed

rtl/verilog/mor1kx_execute_alu.v
stffrdhrn: Which check fails? Division is differnt from multiplication right?

stffrdhrn:
> Added checks for the serial features like multiplier, divider, and shifter. Let me know if the approach used is correct.
> We need to update the SBY file when the feature type is changed. Shall I add tasks for all features? This may create too many `execute_alu` directories after the test.

I think adding the tasks is good if it all passes. The extra directories is not an issue, how much more time will it take?

rtl/verilog/mor1kx_execute_alu.v
Harshitha172000: Sorry, it's a typo mistake. Actually, it's possible to have decoded instruction of multiplication type even after setting multiplication feature as none. This parameter is something that we set externally but the instruction can be anything. If the solver considers opcode for multiplication and we have set `FEATURE_MULTIPLIER == "NONE"`, then the test fails.

Harshitha172000:
> I think adding the tasks is good if it all passes. The extra directories is not an issue, how much more time will it take?

All tests pass and I have added four tasks. Tags just specify parameters that are set. I tried using tags in pycode to set parameters but yosys shows an error while decoding string parameters, so not using pycode. 
The entire alu test may take one or two minutes.

stffrdhrn (commented):
Comments on the serial ALU operations. I think we can change how we check this.  I think its better to not check exact numbers, just check bounds.

rtl/verilog/mor1kx_fetch_cappuccino.v
```diff
@@ -642,4 +642,200 @@ end else begin
...
+   end
+
+//------------------FETCH PROPERTIES----------------
+
+   //Fetch valid stays high for one clock cycle if there is no stall in pipeline
+   always @(posedge clk `OR_ASYNC_RST) begin
+      if (f_past_valid && $past(fetch_valid_o) && $past(mispredict_stall)
+          && !$past(stall_fetch_valid))
+         assert (!fetch_valid_o);
+   end
```
stffrdhrn, line 683: I don't really get this.  Maybe the comment is missing something.  The test is during the case that there was a `mispredict_stall`.   In general `fetch_valid_o` should be true all the time unless there is a stall due to ICACHE misses.

Maybe the comment should be something like
`// If fetch was valid it will be cleared when we have a branch misprediction`

```
   // when stalled, we need to maintain fetch_valid_o
   assign stall_fetch_valid = !padv_i & fetch_valid_o;

   // fetch_valid_o generation
   always @(posedge clk `OR_ASYNC_RST)
     if (rst)
       fetch_valid_o <= 1'b0;
     else if (pipeline_flush_i)
       fetch_valid_o <= 1'b0;
     else if (bus_access_done & padv_i & !mispredict_stall & !immu_busy &
	      !tlb_reload_busy | stall_fetch_valid)
       fetch_valid_o <= 1'b1;
     else
       fetch_valid_o <= 1'b0;
```

rtl/verilog/mor1kx_execute_alu.v
```diff
@@ -851,4 +851,264 @@ endgenerate
...
+          end
+          if (f_mul_count == 30)
+             f_mul <= 0;
+      end
+   end
+
+   always @(posedge clk)
+      if (f_past_valid && !$past(rst) && FEATURE_MULTIPLIER == "SERIAL"
+          && mul_valid && $fell(f_mul))
+          assert(f_mul_count == 6'd31);
```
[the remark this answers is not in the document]
Harshitha172000: `f_mul_count` never goes beyond 32 because the number of bits reserved for it is 5. Its maximum count is 31.

rtl/verilog/mor1kx_execute_alu.v
```diff
@@ -851,4 +851,264 @@ endgenerate
...
+             f_mul <= 1;
+             f_mul_count <= 0;
+          end
+          if (f_mul) begin
+             assert (!mul_valid);
+             f_mul_count <= f_mul_count + 1;
+             `ASSUME (op_mul_i);
+             `ASSUME (decode_valid_i);
+          end
+          if (f_mul_count == 30)
```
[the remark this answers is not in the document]
Harshitha172000: If I add this, the `f_mul` signal fell in the next clock cycle, so we may end up checking assertion in the next clock cycle after we get mul_valid,

rtl/verilog/mor1kx_fetch_cappuccino.v
Harshitha172000: I will update the comment.

rtl/verilog/mor1kx_decode.v
stffrdhrn: it was removed

rtl/verilog/mor1kx_decode.v
stffrdhrn: it got removed

rtl/verilog/mor1kx_decode.v
stffrdhrn: removed

rtl/verilog/mor1kx_execute_alu.v
stffrdhrn: I think so

rtl/verilog/mor1kx_execute_alu.v
stffrdhrn: ok

Harshitha172000:
Modified check for Three-stage multiplier. Previously, input assumption fails during the top module test. So, verification was limited to ALU. 

Before

```
`ifdef ALU
   //Three stage Multiplier takes 3clks to output the result
   always @(posedge clk) begin
      if (FEATURE_MULTIPLIER == "THREESTAGE" &&
          $past(op_mul_i,3) && f_past_valid &&
          !$past(rst) && !$past(rst,2) && !$past(rst,3))
         //Assume multiplication opcode is valid for three clock cycles
         assume ($past(op_mul_i) && $past(op_mul_i,2));
   end
   //MULTIPLICATION: THREE STAGE
   always @(posedge clk) begin
      if (!$past(op_mul_i,4) && $past(op_mul_i,3) &&
          f_past_valid && FEATURE_MULTIPLIER == "THREESTAGE"
          && !$past(rst) && !$past(rst,2) && !$past(rst,3) &&
          mul_valid && $past(decode_valid_i,4) && !$past(rst,4)) begin
         //Stall three clock cycles
         assert ($past(alu_stall) && $past(alu_stall,2) && $past(alu_stall,3));
         //After completion, no stall
         assert (!alu_stall);
         assert (alu_valid_o);
      end
   end
`endif
```

After investigation, I noticed that `pipeline_flush_i` affects input assumption due to which test fails in CPU.

Modified code

```
   //MULTIPLICATION: THREE STAGE
   reg f_op_mul;
   initial f_op_mul = 0;

   always @(posedge clk) begin
      if (FEATURE_MULTIPLIER == "THREESTAGE"
          && f_past_valid && !$past(rst)) begin
          if ($rose(op_mul_i) && !mul_valid && !pipeline_flush_i)
             f_op_mul <= 1;
          if (($rose(f_op_mul) || (!$past(f_op_mul,2) && $past(f_op_mul))) 
              && ((!$past(pipeline_flush_i) && !pipeline_flush_i) ||
              !pipeline_flush_i))
             `ASSUME ($stable(op_mul_i));
          if ((!$past(f_op_mul,2) && $past(f_op_mul) && !$past(rst,2)
              && op_mul_i) || !pipeline_flush_i || !op_mul_i)
             f_op_mul <= 0;
      end
   end

   always @(posedge clk)
      if (FEATURE_MULTIPLIER == "THREESTAGE" && !$past(op_mul_i,4)
          && $past(op_mul_i,3) && mul_valid && $past(decode_valid_i,4) &&
          f_past_valid &&  !$past(rst) && !$past(rst,2) && !$past(rst,3) &&
          !$past(rst,4) && !(op_div_i | op_mul_i | fpu_op_is_arith |
          fpu_op_is_cmp | op_shift_i | op_ffl1_i) && $fell(f_op_mul)) begin
         //Stall three clock cycles
         assert ($past(alu_stall) && $past(alu_stall,2) && $past(alu_stall,3));
         //After execution, no stall
         assert (!alu_stall);
         assert (alu_valid_o);
      end
```
The additional condition has been added to avoid ALU stall after completing execution. These opcode's arrival in the 4th clock cycle may result in `alu_stall` and may delay valid signal.

rtl/verilog/mor1kx_fetch_cappuccino.v
Harshitha172000: Done

rtl/verilog/mor1kx_execute_alu.v
```diff
@@ -851,4 +851,264 @@ endgenerate
...
+   initial f_mul = 0;
+   //Valid multiplication output is seen after 32 clocks
+   always @(posedge clk) begin
+      if (f_past_valid && !$past(rst) && FEATURE_MULTIPLIER == "SERIAL") begin
+          if ($rose(op_mul_i) && decode_valid_i) begin
+             f_mul <= 1;
+             f_mul_count <= 0;
+          end
+          if (f_mul) begin
+             assert (!mul_valid);
```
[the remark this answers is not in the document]
Harshitha172000: Removed

rtl/verilog/mor1kx_execute_alu.v
```diff
@@ -851,4 +851,264 @@ endgenerate
...
+          if ($rose(op_div_i) && decode_valid_i) begin
+             f_div <= 1;
+             f_div_count <= 0;
+          end
+          if (f_div) begin
+             assert (!div_valid);
+             f_div_count <= f_div_count + 1;
+             `ASSUME (decode_valid_i);
+          end
+          if (f_div_count == 30)
```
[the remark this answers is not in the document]
Harshitha172000: Why?

rtl/verilog/mor1kx_execute_alu.v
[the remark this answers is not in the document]
stffrdhrn: I have a fix for this.  Lets merge as is, then we can work on cleaning up some of these things.

Issue #122, opened 2021-06-20
sequential mtspr calls fail


stffrdhrn:
I have converted to or1k-tests to use inline calls to `mtspr()` and `mfspr()`. This causes failures in the following cases in `or1k-mmu`.

See: https://github.com/openrisc/or1k-tests/commit/32f0ba26bd581ca8cf7d3b4c46a53e2509acdcce

```c
static int itlb_match_test (int way, int set)
{
...

  // Flush these areas incase cache doesn't write them through immediately
  mtspr(OR1K_SPR_DCACHE_DCBFR_ADDR, ta - 8);
  mtspr(OR1K_SPR_DCACHE_DCBFR_ADDR, ta);
(CASE 1 - assembly stores ta to stack at this time)

  mtspr(OR1K_SPR_DCACHE_DCBFR_ADDR, ta + PAGE_SIZE - 8);
  mtspr(OR1K_SPR_DCACHE_DCBFR_ADDR, ta + PAGE_SIZE);

  /* Shifting one in ITLBMR */
  add = (PAGE_SIZE*itlb_sets);
  // Space we'll access and expect the MMU to translate our requests
  ea = add + (set*PAGE_SIZE);

  /* Enable IMMU */
  immu_enable ();

  while (add != 0x00000000) {
(CASE 2)
    mtspr (OR1K_SPR_IMMU_ITLBW_MR_ADDR(way, set), ea | OR1K_SPR_IMMU_ITLBW_MR_V_MASK);
    mtspr (OR1K_SPR_IMMU_ITLBW_TR_ADDR(way, set), ta | ITLB_PR_NOLIMIT);

```
## Case 1 - this happens when 1 nop after mtspr.
The first case is where we have `mtspr` following by a `l.sw`, this causes an invalidate to be sent to the dcache.  At the same time as the write operation.  The write fails.

## Case 2 - this happens with 0 nop after mtspr.
The mtspr followed by mtspr causes the second write to OR1K_SPR_IMMU_ITLBW_TR_ADDR to be lost.

## Workaround
Adding 2 `l.nop`s after `mtspr` fixes the issue.   Below in case 1 the `l.sw` came before the `l.nop` due to some optimizations from gcc, to fix it the `asm` for `l.mtspr` contains the nop inline i.e. `l.mtspr %0, %1, 0; l.nop; l.nop`.  But the below is without the fix exposing the CPU issue, but it should still work.

## Trace case1 - dcache spr mix with l.sw fails

```
[...]

PR #124, opened 2021-06-24
Fix issues related to successive mtspr/mtspr instructions

stffrdhrn:
Issue #122

Writing to the tlb rams was dependent on the spr_bus_ack as a means
to only write during the first cycle after a spr_bus_stb_i.  This works
fine for a single write, but if we have a write to itlb_match_spr
followed by itlb_trans_spr the spr_bus_ack never goes low so the write
to itlb_trans_spr is skipped.

Fix this by using the itlb_match_spr_cs_r and itlb_trans_spr_cs_r
signals to control the "first cycle after spr_bus_stb_i" condition. This
way writes can happen independently.

Note, there may still be issues with a write to itlb_match followed by
another write to itlb_match.  Fixing that requires changing the "first
cycle after spr_bus_stb_i" policy, which is similar to dmmu.

### DIFF
--- a/rtl/verilog/mor1kx_dcache.v
+++ b/rtl/verilog/mor1kx_dcache.v
@@ -342,6 +342,9 @@ module mor1kx_dcache
     */
    integer w1;
    always @(posedge clk `OR_ASYNC_RST) begin
+      // The default is (of course) not to acknowledge the invalidate
+      invalidate_ack <= 1'b0;
+
       if (rst) begin
 	 state <= IDLE;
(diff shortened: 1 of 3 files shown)


stffrdhrn:
This should fix all of #122 now as the test case passes for me know.  This also fixes a trivial issue found related to cache hits not being forwarded to the performance counters.

stffrdhrn:
If anyone has time please help review @olofk @bandvig @wallento @juliusbaxter @skristiansson, it does pass tests.

PR #131, opened 2021-07-18
formal: Update ALU serial operations to time until complete

stffrdhrn:
The previous verification logic counted the cycles it took
to run a multipy operation stopping when it reached 30. then
confirming we were done.  The new logic is a bit more flexible and
simple we count until the operation is complete.  If the count goes
above 32 we fail.  This can be used for any multi clock multiply
operation so we use for 3 stage as well.

In the .sby file I have updated depth to 64 to allow k-indection to
capture a valid start of a multiply operation.  Also, I have changed
task names to be a bit more descriptive.

### DIFF
--- a/bench/formal/f_multiclock_op.v
+++ b/bench/formal/f_multiclock_op.v
@@ -1,0 +1,66 @@
+/* ****************************************************************************
+  This Source Code Form is subject to the terms of the
+  Open Hardware Description License, v. 1.0. If a copy
+  of the OHDL was not distributed with this file, You
+  can obtain one at http://juliusbaxter.net/ohdl/ohdl.txt
+
+  Description: mor1kx formal multiclock alu operation checker
+
+  Checks that an operation f_op_i finishes within a number of clock
+  cycles.  The completion of the operation is signaled by asserting
+  f_op_valid_i.  This is used to validate multi operation ALU operations
+  like multiply, divide and shift.
+
+  Copyright (C) 2021 Stafford Horne <shorne@gmail.com>
+
+***************************************************************************** */
+
+module f_multiclock_op
+#(
+  parameter STABLE_WIDTH = 32,
+  parameter OP_MAX_CLOCKS = 32
+ )
+ (
+  input clk,
+  input f_op_i,
+  input f_op_valid_i,
+  input decode_valid_i,
+  input [STABLE_WIDTH-1:0] f_stable_i,
+  input f_past_valid
+);
+
+   reg [5:0] f_op_count;
+   reg f_op;
+   initial f_op_count = 0;
+   initial f_op = 0;
+
+   //Valid multiplication output is seen after 32 clocks
+   always @(posedge clk) begin
+      if (f_past_valid) begin
+	  // If we have slow operations decode_valid will
+	  // go low to stall after the operation starts.
+	  if ($past(f_op_i))
+	     assume (!decode_valid_i);
+	  else
+	     assume (decode_valid_i);
+
+	  // Start counting when we see the operation start
+	  if ($rose(f_op_i)) begin
+	     f_op <= 1;
+	     f_op_count <= 1;
+	  end else if (f_op_valid_i) // Stop when the op is done
+	     f_op <= 0;
+
+	  if (f_op) begin
+	     f_op_count <= f_op_count + 1;
+	     assume (f_op_i);
+	     assume ($stable(f_stable_i));
+	  end
+
+	  // Ensure the operation completes never going beyond max count
+	  assert (f_op_count <= (OP_MAX_CLOCKS + 2));
+      end
+   end
+
+endmodule
+
--- a/bench/formal/mor1kx_execute_alu.sby
+++ b/bench/formal/mor1kx_execute_alu.sby
@@ -1,35 +1,41 @@
 [tasks]
 #Tags are not used in code but specify type of parameters set.
-test1  serial_mul        serial_div       serial_shift
-test2  pipelined_mul     simulation_div   barrel_shift
-test3  threestage_mul    none  _div       serial_div
-test4  simulation_mul    simulation_div   barrel_shift
+test_serial      serial_mul        serial_div       serial_shift
+test_pipe_mul    pipelined_mul     simulation_div   barrel_shift
+test_3stage_mul  threestage_mul    no_div           serial_shift
+test_sim_mul     simulation_mul    simulation_div   barrel_shift
 
+# Depth set to 64 to handle 32 cycle serial operations
+# For other tests no operations take more than 3 cycles so set
+# the depth to a conservative 10
 [options]
 mode prove
-depth 20
+depth 10
+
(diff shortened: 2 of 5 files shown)


rtl/verilog/mor1kx_execute_alu.v
```diff
@@ -946,34 +915,42 @@ endgenerate
...
+   wire [5:0] f_mul_max_clks;
+   initial f_mul_count = 0;
    initial f_mul = 0;
+   assign f_mul_max_clks = FEATURE_MULTIPLIER == "SERIAL" ? 6'd32 : 6'd3;
+
+   always @(posedge clk)
+      if ($past(op_mul_i))
+        `ASSUME (!decode_valid_i);
+      else if (op_mul_i && f_past_valid)
+        `ASSUME (decode_valid_i);
```
stffrdhrn, line 930: This probably should be either 1, more generic to handle any multi cycle statement or 2. moved into the SERIAL/THREESTAGE if statement internally.

rtl/verilog/mor1kx_execute_alu.v
```diff
@@ -946,34 +915,42 @@ endgenerate
...
              f_mul_count <= f_mul_count + 1;
              `ASSUME (op_mul_i);
-             `ASSUME (decode_valid_i);
+             `ASSUME ($stable(rfa_i));
+             `ASSUME ($stable(rfb_i));
+             `ASSUME ($stable(op_mul_signed_i));
           end
-          if (f_mul_count == 30)
-             f_mul <= 0;
+          assert (genblk2.genblk1.serialmultiply.serial_mul_cnt <= 6'b10_0000);
```
stffrdhrn, line 949: This asserts an internal signal, to ensure k-indction doesn't start out setting serial_mul_cnt really high which is not possible.

rtl/verilog/mor1kx_execute_alu.v
stffrdhrn: This also causes the mor1kx_cpu verification to fail

stffrdhrn: I fixed the first issue.

rtl/verilog/mor1kx_execute_alu.v
Harshitha172000: I think this assertion is not checked in the CPU test.

image

bench/formal/mor1kx_execute_alu.sby
```diff
@@ -1,35 +1,36 @@
...
+test_serial      serial_mul        serial_div       serial_shift
+test_pipe_mul    pipelined_mul     simulation_div   barrel_shift
+test_3stage_mul  threestage_mul    no_div           serial_shift
+test_sim_mul     simulation_mul    simulation_div   barrel_shift
 
+# Depth set to 64 to handle 32 cycle serial operations
 [options]
 mode prove
-depth 20
+depth 64
```
Harshitha172000, line 11: Maybe we can add depth 64 only for test_serial.  For other tasks, we can reduce depth.

rtl/verilog/mor1kx_execute_alu.v
Harshitha172000: > This also causes the mor1kx_cpu verification to fail

`pipeline_flush_i` stops ongoing execution due to which `op_mul_i` may not be high for 3 or 32 clks. So, the assetion fails.

stffrdhrn: We can add pipeline flush to the logic.

rtl/verilog/mor1kx_execute_alu.v
stffrdhrn: Yeah. I am working on a fix for it

Harshitha172000: By default, the THREE STAGE multiplier is checked. If we change it to SERIAL we may not get this warning.

rtl/verilog/mor1kx_execute_alu.v
```diff
@@ -946,34 +915,42 @@ endgenerate
...
+          assert (f_mul_count <= (f_mul_max_clks + 2));
       end
```
Harshitha172000, line 951: As per new logic, `f_mul` fall after `mul_valid` arrives. But count continues when `mul_valid` is high and may assume `op_mul_i` and other assumptions in the 33rd clock cycle which may fail in the top module test.

image

Adding `!mul_valid` in condition may not do extra assumptions.

```
          if (f_mul & !mul_valid) begin
             f_mul_count <= f_mul_count + 1;
             `ASSUME (op_mul_i);
             `ASSUME ($stable(rfa_i));
             `ASSUME ($stable(rfb_i));
             `ASSUME ($stable(op_mul_signed_i));
          end
assert (genblk2.genblk1.serialmultiply.serial_mul_cnt <= 6'b10_0000);
assert (f_mul_count <= (f_mul_max_clks + 1));
```

stffrdhrn:
I converted the multiclock check to a module so we can reuse the logic.  Also the CPU level checks are fixed.

bench/formal/mor1kx_execute_alu.sby
stffrdhrn: Good idea I didn't know about that option

rtl/verilog/mor1kx_execute_alu.v
stffrdhrn: For CPU I changed the multiplier to SIMULATION that way we don't need to check multiclock as we already verified it in the execute alu module test.

stffrdhrn:
I squashed some changes and rebased it on master after merging the formal-CI PR.

rtl/verilog/mor1kx_execute_alu.v
stffrdhrn: Thats a good point. Its why I have the +2 not +1, I can add the extra check now to `f_multiclock_op` module.

bench/formal/f_multiclock_op.v
```diff
@@ -0,0 +1,66 @@
...
+   initial f_op_count = 0;
+   initial f_op = 0;
+
+   //Valid multiplication output is seen after 32 clocks
+   always @(posedge clk) begin
+      if (f_past_valid) begin
+	  // If we have slow operations decode_valid will
+	  // go low to stall after the operation starts.
+	  if ($past(f_op_i))
+	     assume (!decode_valid_i);
```
Harshitha172000, line 43: It's better to turn assume to assert during cpu test which ensures that our assumptions are valid.

Harshitha172000:
We can merge this PR for now, and can check cpu test after adding input assumptions to mor1kx_cpu_cappuccino.

bench/formal/f_multiclock_op.v
stffrdhrn: Yes, but for now I am not using this during cpu tests.  We can add it later.

PR #132, opened 2021-07-19
Formal: Update ALU Multiclock Interface

Harshitha172000:
Previously, the test passed for serial checks of depth 10. We can correctly
validate serial operations only if the solver takes steps more than 40, and proof
should fail if we give depth less than OP_MAX_CLOCKS. Changing assumption
condition on decode_valid_i fixed this.

Also, Multi clock Interface has been used to verify FFL1.

### DIFF
--- a/bench/formal/f_multiclock_op.v
+++ b/bench/formal/f_multiclock_op.v
@@ -39,7 +39,7 @@ module f_multiclock_op
       if (f_past_valid) begin
 	  // If we have slow operations decode_valid will
 	  // go low to stall after the operation starts.
-	  if ($past(f_op_i))
+	  if (f_op | f_op_valid_i)
 	     assume (!decode_valid_i);
 	  else
 	     assume (decode_valid_i);
@@ -58,9 +58,8 @@ module f_multiclock_op
 	  end
 
 	  // Ensure the operation completes never going beyond max count
-	  assert (f_op_count <= (OP_MAX_CLOCKS + 2));
+	  assert (f_op_count <= (OP_MAX_CLOCKS + 1));
       end
    end
-
 endmodule
 
--- a/bench/formal/mor1kx_execute_alu.sby
+++ b/bench/formal/mor1kx_execute_alu.sby
@@ -2,7 +2,7 @@
 #Tags are not used in code but specify type of parameters set.
 test_serial      serial_mul        serial_div       serial_shift
 test_pipe_mul    pipelined_mul     simulation_div   barrel_shift
-test_3stage_mul  threestage_mul    no_div           serial_shift
+test_3stage_mul  threestage_mul    no_div           barrel_shift
 test_sim_mul     simulation_mul    simulation_div   barrel_shift
 
 # Depth set to 64 to handle 32 cycle serial operations
@@ -31,13 +31,14 @@ test_pipe_mul: chparam -set OPTION_SHIFTER "BARREL" mor1kx_execute_alu
 
 test_3stage_mul: chparam -set FEATURE_MULTIPLIER "THREESTAGE" mor1kx_execute_alu
 test_3stage_mul: chparam -set FEATURE_DIVIDER "NONE" mor1kx_execute_alu
-test_3stage_mul: chparam -set OPTION_SHIFTER "SERIAL" mor1kx_execute_alu
+test_3stage_mul: chparam -set OPTION_SHIFTER "BARREL" mor1kx_execute_alu
 
 test_sim_mul: chparam -set FEATURE_MULTIPLIER "SIMULATION" mor1kx_execute_alu
 test_sim_mul: chparam -set FEATURE_DIVIDER "SIMULATION" mor1kx_execute_alu
 test_sim_mul: chparam -set OPTION_SHIFTER "BARREL" mor1kx_execute_alu
 
 chparam -set FEATURE_OVERFLOW "ENABLED" mor1kx_execute_alu
+chparam -set FEATURE_FFL1 "REGISTERED" mor1kx_execute_alu
 
 
 prep -top mor1kx_execute_alu
--- a/rtl/verilog/mor1kx_execute_alu.v
+++ b/rtl/verilog/mor1kx_execute_alu.v
@@ -889,7 +889,7 @@ endgenerate
          `ASSUME (op_div_signed_i | op_div_unsigned_i);
 
    always @(posedge clk)
-      if (f_past_valid && !$past(rst))
+      if (f_past_valid)
          `ASSUME ($onehot0(f_exec_opcodes));
 
    //MULTIPLICATION: PIPELINED
@@ -907,7 +907,7 @@ endgenerate
    //MULTIPLICATION: SIMULATION
    always @(*) begin
       if (FEATURE_MULTIPLIER == "SIMULATION" &&
-          op_mul_i) begin
+          op_mul_i && !rst) begin
          //No stall for simulation based multiplier
          assert (!alu_stall);
          assert (alu_valid_o);
@@ -941,7 +941,7 @@ endgenerate
    end
 
    //TODO
-   //Formal verification of FPU and FF1.
+   //Formal verification of FPU.
 
    //DIVIDER: SIMULATION
    always @(posedge clk) begin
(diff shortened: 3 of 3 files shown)


bench/formal/mor1kx_cpu_cappuccino.sby
```diff
@@ -39,16 +39,18 @@ read -formal mor1kx_bus_if_wb32.v
...
 read_verilog mor1kx_wb_mux_cappuccino.v
 read_verilog mor1kx_cfgrs.v
+read -formal -DCPU f_multiclock_op.v
 
 chparam -set FEATURE_INSTRUCTIONCACHE "ENABLED" mor1kx_cpu_cappuccino
 chparam -set FEATURE_IMMU "ENABLED" mor1kx_cpu_cappuccino
 chparam -set FEATURE_DMMU "ENABLED" mor1kx_cpu_cappuccino
 chparam -set FEATURE_DATACACHE "ENABLED" mor1kx_cpu_cappuccino
-chparam -set FEATURE_MULTIPLIER "SIMULATION" mor1kx_cpu_cappuccino
+chparam -set FEATURE_MULTIPLIER "SERIAL" mor1kx_cpu_cappuccino
```
stffrdhrn, line 48: We will still only run for 5 clocks?

bench/formal/f_multiclock_op.v
```diff
@@ -26,39 +26,49 @@ module f_multiclock_op
...
+            f_op_count <= f_op_count + 1;
+            `ASSUME (!decode_valid_i);
+            `ASSUME (f_op_i);
+            `ASSUME ($stable(f_stable_i));
+         end
 
-	  // Ensure the operation completes never going beyond max count
-	  assert (f_op_count <= (OP_MAX_CLOCKS + 2));
+         // Ensure the operation completes never going beyond max count
+         if ($fell(f_op))
```
stffrdhrn, line 70: I don't eat too tech this after the fall, but always. Otherwise we get a pass if f_op_i is always high and the instruction runs forever.

bench/formal/f_multiclock_op.v
```diff
@@ -26,39 +26,49 @@ module f_multiclock_op
...
 
-	  if (f_op) begin
-	     f_op_count <= f_op_count + 1;
-	     assume (f_op_i);
-	     assume ($stable(f_stable_i));
-	  end
+         // Count number of clock cycles taken to complete op
+         if (f_op & f_op_i & !f_op_valid_i) begin
+            f_op_count <= f_op_count + 1;
+            `ASSUME (!decode_valid_i);
```
stffrdhrn, line 64: I don't like mixing the decode valid assumptions into here. It makes it hard to understand.

bench/formal/mor1kx_execute_alu.sby
```diff
@@ -12,8 +12,8 @@ test_sim_mul     simulation_mul    simulation_div   barrel_shift
 mode prove
 depth 10
 
-test_serial: depth 64
-
+test_serial: depth 40
+test_3stage_mul: depth 40
```
Harshitha172000, line 16: I think task `test_3stage_mul` requires more depth as it verifies the serial shifter.

Harshitha172000:
Only the `OP_MAX_CLOCKS` assertion needs more depth in the top module test but not input assumption validation. I have verified all input assumptions with the added constraints and it passes the test.

rtl/verilog/mor1kx_execute_alu.v
```diff
@@ -1006,8 +1010,29 @@ endgenerate
...
+
+   //FFL1
+   generate
+      if (FEATURE_FFL1 == "REGISTERED") begin : f_ffl1_multiclock
+          f_multiclock_op #(
+             .STABLE_WIDTH(`OR1K_ALU_OPC_WIDTH),
+             .OP_MAX_CLOCKS(1)
+           ) u_f_multiclock_ffl1 (
+             .clk(clk),
+             .f_op_i(op_ffl1_i),
```
Harshitha172000, line 1029: I have used multi clock interface to check instructions l.ff1 or l.fl1. This check is applicable only to registered FFL1 as there is one clock cycle delay. For non-registered FFL1 implementation, the entire logic is driven by mux. So, not asserting it.

bench/formal/f_multiclock_op.v
```diff
@@ -26,41 +26,74 @@ module f_multiclock_op
...
+            // Confirm no discontinuity in op_i signal while executing op, given
+            // pipeline flush won't occur.
+            `ASSUME (f_op_i);
+            `ASSUME ($stable(f_stable_i));
+         end
+`ifdef ALU
+         // Ensure the operation completes never going beyond max count
+         // This assertion check needs depth of 40 or more.
+         assert (f_op_count <= OP_MAX_CLOCKS + 1);
+`endif
```
stffrdhrn, line 78: why do we do this only for ALU?

bench/formal/f_multiclock_op.v
```diff
@@ -26,41 +26,74 @@ module f_multiclock_op
...
+         // Start counting when we see the operation start and the previous
+         // execution op is completed.
+         if ($rose(f_op_i) & alu_valid_o) begin
+            f_op <= 1;
+            f_op_count <= 1;
+         end
+         // Stop when op is done or abort if op_i goes low during execution.
+         else if (f_op_valid_i | !f_op_i) begin
+            f_op <= 0;
+         end
```
stffrdhrn, line 64: Should we stop if valid or flush?

rtl/verilog/mor1kx_execute_alu.v
stffrdhrn: ok

stffrdhrn (changes requested):
Overall I think this is not adding much value to the original implementation.  Now that we are not enabling this for CPU level tests not much of the change is needed.

bench/formal/f_multiclock_op.v
Harshitha172000: When we get valid, then the operation is completed, so stop counting. Ongoing op will be incomplete if `f_op_i` goes low in between; this happens when `pipeline_flush_i` arrives. Deasserting `f_op` enables us to start counting from zero when we get a new op, or else the count may increment from the old value.

bench/formal/f_multiclock_op.v
Harshitha172000: The proof will slow down if we set a depth of 40 for the CPU test.

Harshitha172000:
> Overall I think this is not adding much value to the original implementation. Now that we are not enabling this for CPU level tests not much of the change is needed.

I just tried to validate input assumptions from the CPU's point of view. Now, I think it's not necessary to validate assumptions in the interface file. I will revert these changes.

bench/formal/f_multiclock_op.v
```diff
@@ -26,41 +26,74 @@ module f_multiclock_op
...
+ // Below assertions can only be true if f_op_count is equal to OP_MAX_CLOCKS,
+ // so restricting test only to ALU.
+`ifdef ALU
+   always @(posedge clk)
+      if ((f_op_count == OP_MAX_CLOCKS) & $changed(f_op_count) & f_past_valid &
+          !pipeline_flush_i) begin
+        // op is completed when count reaches OP_MAX_CLOCKS where valid signals
+        // are expected to be high.
+        assert ($past(f_op_valid_i));
+        assert ($past(alu_valid_o));
```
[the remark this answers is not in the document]
Harshitha172000: ok

rtl/verilog/mor1kx_execute_alu.v
```diff
@@ -907,7 +907,7 @@ endgenerate
    //MULTIPLICATION: SIMULATION
    always @(*) begin
       if (FEATURE_MULTIPLIER == "SIMULATION" &&
-          op_mul_i) begin
+          op_mul_i & !op_ffl1_i) begin
```
stffrdhrn, line 910: Why do we need this `!op_ffl1_i` condition?  Shouldnt the op be part of the one hot assumption?

rtl/verilog/mor1kx_execute_alu.v
```diff
@@ -1006,7 +1006,24 @@ endgenerate
              .f_op_i(op_shift_i),
              .f_op_valid_i(shift_valid),
              .decode_valid_i(decode_valid_i),
-             .f_stable_i({rfa_i, rfb_i, opc_alu_secondary_i}),
+             .f_stable_i({a, b, opc_alu_secondary_i}),
```
stffrdhrn, line 1009: Why change from `rfa_i` to `a` (similar for `b`)?

rtl/verilog/mor1kx_execute_alu.v
Harshitha172000: The one-hot assumption is not considered when reset is high, so this assertion fails when we simultaneously have `op_mul_i` and `op_ffl1_i` high. Anyhow, I will remove 'op_ffl1_i' here and modify one_hot.

rtl/verilog/mor1kx_execute_alu.v
Harshitha172000: Because the number of shifts may be specified either through register or immediate.

stffrdhrn: I dont underrstand what you mean,  Can you change it back to `rfa_i, rfb_i, opc_alu_secondary_i`?  If you need other things to be stable you can add those as well.

Harshitha172000:
```if (CALCULATE_BRANCH_DEST=="TRUE") begin : calculate_branch_dest
   assign a = (op_jbr_i | op_jr_i) ? pc_execute_i : rfa_i;
   assign b = immediate_sel_i ? immediate_i :
              op_jbr_i ? {{4{immjbr_upper_i[9]}},immjbr_upper_i,imm16_i,2'b00} :
              rfb_i;
end else begin
   assign a = rfa_i;
   assign b = immediate_sel_i ? immediate_i : rfb_i;
```

Serial shifter shifts `a` by `b` number of bits. `b` value may be stored in the register, or it can be an immediate value.
Instead of including all immediate inputs `immjbr_upper_i, immediate_sel_i, immediate_i` , I kept `b` stable.

bench/formal/mor1kx_execute_alu.sby
Harshitha172000: Fine. I will add a barrel shifter for `test_3stage_mul`

stffrdhrn (approved):

Issue #127, opened 2021-07-08
spr_bus_ack_o reset issue


Harshitha172000:
This issue is applicable to the mor1kx_icache module. After resetting, the state moves to IDLE but spr_bus_ack_o can be high if spr strobe/write signals stays high during reset. This is possible as spr stb/we signals are not reset in the ctrl module.

Assertion Failed:
```
          always @(posedge clk)
               if ($past(rst) && f_past_valid)
                  assert (!spr_bus_ack_ic_i);
```
Trace showing issue:

![image](https://user-images.githubusercontent.com/59602295/124925413-086c0300-e01a-11eb-8dba-0f704c53dc2f.png)

stffrdhrn:
I posted a fix for this does it also fix formal?

Harshitha172000:
No, now I see this trace.  With the reset, the state switches from INVALIDATE to IDLE and keeps spr_bus_ack_o high.

![image](https://user-images.githubusercontent.com/59602295/125103226-c4502f80-e0f9-11eb-8100-06869bdd5d00.png)

stffrdhrn:
OK, it doesn't cater for the case when `spr_bus_ack_o` is high to start with.

stffrdhrn:
I pushed another update, how does it look now?

Harshitha172000:
It passes the test, the issue is resolved.

PR #133, opened 2021-08-01
Formal: Wishbone Interface

Harshitha172000:
Formally verifying properties of "B3_READ_BURSTING"/"B3_REGISTERED_FEEDBACK"  wishbone. Property checks for IBUS_BRIDGE pass, but DBUS_BRIDGE fails in some cases. "CLASSIC" Wishbone is not reset, so many properties of this interface are not satisfied. The maximum request value is set to 16, which can be increased but should choose 'F_LGDEPTH' carefully, or else the test won't pass `f_outstanding` validations.

### DIFF
--- a/bench/formal/fwb_master.v
+++ b/bench/formal/fwb_master.v
@@ -1,0 +1,442 @@
+////////////////////////////////////////////////////////////////////////////////
+// }}}
+// Copyright (C) 2017-2021, Gisselquist Technology, LLC
+// {{{
+// This file is part of the WB2AXIP project.
+//
+// The WB2AXIP project contains free software and gateware, licensed under the
+// Apache License, Version 2.0 (the "License").  You may not use this project,
+// or this file, except in compliance with the License.  You may obtain a copy
+// of the License at
+//
+//	http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+// WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+// License for the specific language governing permissions and limitations
+// under the License.
+//
+////////////////////////////////////////////////////////////////////////////////
+//
+//
+`default_nettype none
+// }}}
+module	fwb_master #(
+		// {{{
+		parameter		AW=32, DW=32,
+					F_MAX_ACK_DELAY = 0,
+		parameter		F_LGDEPTH = 4,
+		parameter [(F_LGDEPTH-1):0] F_MAX_REQUESTS = 0,
+		// OPT_BUS_ABORT: If true, the master can drop CYC at any time
+		// and must drop CYC following any bus error
+		parameter [0:0]		OPT_BUS_ABORT = 1'b1,
+		//
+		// If true, allow the bus to be kept open when there are no
+		// outstanding requests.  This is useful for any master that
+		// might execute a read modify write cycle, such as an atomic
+		// add.
+		parameter [0:0]		F_OPT_RMW_BUS_OPTION = 1,
+		//
+		//
+		// If true, allow the bus to issue multiple discontinuous
+		// requests.
+		// Unlike F_OPT_RMW_BUS_OPTION, these requests may be issued
+		// while other requests are outstanding
+		parameter	[0:0]	F_OPT_DISCONTINUOUS = 1,
+		//
+		//
+		// If true, insist that there be a minimum of a single clock
+		// delay between request and response.  This defaults to off
+		// since the wishbone specification specifically doesn't
+		// require this.  However, some interfaces do, so we allow it
+		// as an option here.
+		parameter	[0:0]	F_OPT_MINCLOCK_DELAY = 0,
+		//
+		//
+		//
+		localparam [(F_LGDEPTH-1):0] MAX_OUTSTANDING
+						= {(F_LGDEPTH){1'b1}},
+		localparam	DLYBITS= (F_MAX_ACK_DELAY < 4) ? 2
+				: (F_MAX_ACK_DELAY >= 65536) ? 32
+				: $clog2(F_MAX_ACK_DELAY+1),
+		//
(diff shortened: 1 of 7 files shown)


rtl/verilog/mor1kx_bus_if_wb32.v
```diff
@@ -193,43 +193,50 @@ module mor1kx_bus_if_wb32
...
+   //  000- classic
    always @*
       assert (wbm_cti_o == 3'b111 | wbm_cti_o == 3'b010 | wbm_cti_o == 3'b000);
 
-   //If cycle type identifier is incrementing and wb ack arrives, then burst should be ended.
+   // If cycle type identifier is incrementing and wb ack arrives,
+   // then burst should be ended.
    always @(posedge clk)
-      if (f_past_valid && !$past(rst) && $past(wbm_ack_i) && $past(wbm_cti_o) == 2'b010)
+      if (f_past_valid & !$past(rst) & $past(wbm_ack_i) & (!wbm_ack_i &
```
Harshitha172000, line 237: Correcting erroneous assertion, which holds for any type of wishbone.

bench/formal/fwb_master.v
```diff
@@ -0,0 +1,366 @@
...
+	if ((f_past_valid)&&($past(i_wb_stb))&&(i_wb_stb))
+		`SLAVE_ASSUME(i_wb_we == $past(i_wb_we));
+
+
+	// Within any given bus cycle, the direction may *only* change when
+	// there are no further outstanding requests.
+	always @(posedge i_clk)
+	if ((f_past_valid)&&(f_outstanding > 0))
+		`SLAVE_ASSUME(i_wb_we == $past(i_wb_we));
+
```
Harshitha172000, line 163: The above properties fail for the data bus. The direction of the write enable changes within a series of strobe signals. This happens because `dbus_we_o` won't stay high as it's affected by the `dbus_atomic` signal.


image

bench/formal/fwb_master.v
```diff
@@ -0,0 +1,366 @@
...
+	// Write requests must also set one (or more) of i_wb_sel
+	always @(*)
+	 if ((i_wb_stb)&&(i_wb_we)&f_past_valid)
+		`SLAVE_ASSUME(|i_wb_sel);
```
Harshitha172000, line 167: This property easily passes for ibus since `i_wb_sel` is set as `4'b1111` but for dbus_bridge `i_wb_sel` is equal to `dbus_sel_o`. So, the solver picks up zero value for `i_wb_sel`.

stffrdhrn (approved):
I think overall this looks good, the issues you highlight look reasonable.  I think its better that you add a comment with an issue number i.e. #133 Fails because dbus interface xxx

Then we can open an issue to explain the issue and track the fix.  Could you do this for all of the places you comment out assertions?

rtl/verilog/mor1kx_ctrl_cappuccino.v
```diff
@@ -1642,12 +1642,14 @@ endgenerate
...
+   enabled.
+   always @*
+      assert ($onehot0({spr_read_access, spr_write_access}));
 
-   //always @*
-     //if (spr_bus_stb_o && spr_read_access)
-         //assert (!spr_bus_we_o);
+   always @*
+     if (spr_bus_stb_o && spr_read_access)
+         assert (!spr_bus_we_o);*/
```
Harshitha172000, line 1652: Debug Interface Issue.

rtl/verilog/mor1kx_cpu_cappuccino.v
```diff
@@ -1608,7 +1608,7 @@ module mor1kx_cpu_cappuccino
          assert (!ibus_req_o);
          assert (!spr_bus_stb_o);
          assert (!ibus_burst_o);
-        //Fail assert (!spr_bus_ack_ic_i);
+         assert (!spr_bus_ack_ic_i);
          assert (!spr_bus_ack_immu_i);
```
Harshitha172000, line 1612: Issue #127 has been fixed.

stffrdhrn:
Please resolve conflicts

bench/formal/fwb_master.v
```diff
@@ -0,0 +1,366 @@
...
+// for more details.
+//
+// You should have received a copy of the GNU General Public License along
+// with this program.  (It's in the $(ROOT)/doc directory.  Run make with no
+// target there if the PDF file isn't present.)  If not, see
+// <http://www.gnu.org/licenses/> for a copy.
+//
+// License:	GPL, v3, as defined and found on www.gnu.org,
+//		http://www.gnu.org/licenses/gpl.html
+//
```
stffrdhrn, line 22: Actually, are we allowed to add this to our project?  Dan uses GPL which I think would mean we need to make all our code as GPL to add this.  We are in the process of migrating to CERL OHL-W license https://cern-ohl.web.cern.ch/home

Harshitha172000: I am not sure. Are there any means to include this without altering the license to GPL?

juliusbaxter: First of all, amazing work so far @Harshitha172000 !

This is an interesting question. From my point of view, the verification infrastructure isn't part of what's distributed (ie. the bitstream or the GDS or the physical chip), so actually what you use there doesn't matter, does it? Happy to be convinced otherwise.

Is there a way to make this an external dependency via Fusesoc?

Harshitha172000: Thanks @juliusbaxter

It's a valid point, as this code is not part of the physical chip. I have mailed Dan regarding this. If we can't use it without GPL, then it would be better to move via Fusesoc.

Harshitha172000: I have used another set of wishbone properties identical to the previous one, but now it's licensed under Apache 2. I think the problem is resolved.

stffrdhrn: Thanks, also @juliusbaxter 's suggestion of putting these property definitions into a fusesoc module is a good idea and something I was thinking of.  The whole point of fusesoc is to avoid having to copy and paste modules into different projects.  This seems like a perfect example.
Though right now we are making some modifications to the module to comment out certain properties.

PR #138, opened 2021-08-31
Update Mor1kx-Formal project progress in Readme

Harshitha172000:
We are introducing Formal Verification to Mor1kx CPU. Currently, Formal methods
are applied only to the Cappuccino pipeline, and assertion checks defined as
job `formal verification` support continuous integration parallel with other tests.


stffrdhrn:
This looks fine, but your commit message title says just `Update Readme`.  Its better to be a bit more descriptive. Remember try to explain what changes are done and why.

stffrdhrn:
Thanks, looks good

PR #142, opened 2021-11-28
licensing: replace OHDL headers with CERN OHL-W v2 headers

juliusbaxter:
Here it is folks, as discussed on the mailing list, the switch from my OHDL licence to CERN-OHL-W-2.0.

Mailing list thread: https://lists.librecores.org/pipermail/openrisc/2021-August/003247.html

Full licence text is here: https://ohwr.org/cern_ohl_w_v2.txt

I've omitted the FPU for now because I believe those authors are not on the list and haven't given their approval in writing yet.

One thing I thought maybe is I should include the SPDX identifier? Let me know if you think I should update the PR to include that in the headers as well.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,370 +1,311 @@
-Open Hardware Description License Version 1.0
-(Based on the MPL 2.0 RC2)
-========================================================
-
-1. Definitions
---------------
-
-1.1. "Contributor"
-    means each individual or legal entity that creates, contributes to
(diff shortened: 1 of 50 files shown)


juliusbaxter:
Ugh, I just noticed I committed under my public work git account.

It probably doesn't matter right?

juliusbaxter:
Yeah let me close this PR and re-issue with the commit from the correct account. Sorry.

PR #143, opened 2021-11-28
Licence switch OHDL to CERN OHL-W V2

juliusbaxter:
Here it is folks, as discussed on the mailing list, the switch from my OHDL licence to CERN-OHL-W-2.0.

(This is the second PR replacing the first which was done under the wrong account.)

Mailing list thread: https://lists.librecores.org/pipermail/openrisc/2021-August/003247.html

Full licence text is here: https://ohwr.org/cern_ohl_w_v2.txt

I've omitted the FPU for now because I believe those authors are not on the list and haven't given their approval in writing yet.

One thing I thought maybe is I should include the SPDX identifier in the headers in addition to the README? Let me know if you think I should update the PR to include that in the headers as well.

### DIFF
--- a/bench/formal/Makefile
+++ b/bench/formal/Makefile
@@ -1,8 +1,12 @@
 ###########################################################
-# This Source Code Form is subject to the terms of the
-# Open Hardware Description License, v. 1.0. If a copy
-# of the OHDL was not distributed with this file, You
-# can obtain one at http://juliusbaxter.net/ohdl/ohdl.txt
+# This source describes Open Hardware and is licensed under the CERN-OHLW v2
+#
+# You may redistribute and modify this documentation and make products
+# using it under the terms of the CERN-OHL-W v2 (https:/cern.ch/cern-ohl).
+# This documentation is distributed WITHOUT ANY EXPRESS OR IMPLIED
+# WARRANTY, INCLUDING OF MERCHANTABILITY, SATISFACTORY QUALITY
+# AND FITNESS FOR A PARTICULAR PURPOSE. Please see the CERN-OHL-W v2
+# for applicable conditions.
 #
 # Description: Simplistic makefile for running formal tests
 #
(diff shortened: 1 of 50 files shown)


bench/formal/Makefile
```diff
@@ -1,8 +1,12 @@
...
-# of the OHDL was not distributed with this file, You
-# can obtain one at http://juliusbaxter.net/ohdl/ohdl.txt
+# This source describes Open Hardware and is licensed under the CERN-OHLW v2
+#
+# You may redistribute and modify this documentation and make products
+# using it under the terms of the CERN-OHL-W v2 (https:/cern.ch/cern-ohl).
+# This documentation is distributed WITHOUT ANY EXPRESS OR IMPLIED
+# WARRANTY, INCLUDING OF MERCHANTABILITY, SATISFACTORY QUALITY
+# AND FITNESS FOR A PARTICULAR PURPOSE. Please see the CERN-OHL-W v2
+# for applicable conditions.
```
stffrdhrn, line 9: Rather than putting the full text in each heading should we just put the SPX identifier?  This is what they moved to in the linux kernel.  i.e. In the kernel we have:

```
// SPDX-License-Identifier: GPL-2.0-or-later
/*
 * OpenRISC setup.c
 *
 * Linux architectural port borrowing liberally from similar works of
 ...
```

https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/arch/openrisc/kernel/setup.c?h=v5.16-rc2#n1

stffrdhrn:
Thanks for getting this done @juliusbaxter.

bench/formal/Makefile
juliusbaxter: Yeah I like this. Let me rework the PR. Hopefully won't take me another 3 months to get around to it 🙈

stffrdhrn: Hi @juliusbaxter , did you ever get around to updating the PR?

stffrdhrn: I did it.

PR #148, opened 2022-02-26
Make FPCSR is R/W accessible for both user- and supervisor- modes.

stffrdhrn:


stffrdhrn:
As per new arch spec 1.4 this looks ok.  Note for reads mor1kx already does not protect reading FPCSR

bandvig (approved):
@stffrdhrn 
Merge it, Stafford

PR #154, opened 2023-09-09
ci: Get working on latest ubuntu

stffrdhrn:


stffrdhrn:
formal is failing because the core has issues, this PR is good as it fixes CI

PR #164, opened 2025-03-17
Formal verification for pfpu32_cmp

liujed:
This adds formal verification for the comparison unit in the FPU. Comparison operations don't appear to be fully specified in the ISA spec, but I took my best stab at sensible semantics, based on my understanding of IEEE 754.

For the most part, what I came up with lines up with the implementation. The only difference is the behaviour of `sfult`, `sfule`, `sfugt`, and `sfuge` when given a signalling NaN. I believe these should result in an exception (signalled by `inv_o`), but the implementation doesn't do this.

To get tests to pass, the assertions introduced here have an exception carved out for this behavioural difference. This exception is disabled in the `pfpu32_cmp_snan_bug` test, which currently fails.

### DIFF
--- a/bench/formal/Makefile
+++ b/bench/formal/Makefile
@@ -27,14 +27,16 @@ TESTS := mor1kx_cache_lru \
 	mor1kx_ctrl_cappuccino \
 	mor1kx_rf_cappuccino \
 	mor1kx_bus_if_wb32 \
-	pfpu32_addsub
+	pfpu32_addsub \
+	pfpu32_cmp
 
 # tests that no longer work keep them here so we can easily
 # run them if we want to try and fix them.
 BROKEN_TESTS := \
 	mor1kx_lsu_cappuccino \
 	mor1kx_cpu_cappuccino \
-	mor1kx
+	mor1kx \
+	pfpu32_cmp_snan_bug
 
 all: $(TESTS)
 clean:
--- a/bench/formal/pfpu32_cmp.sby
+++ b/bench/formal/pfpu32_cmp.sby
@@ -1,0 +1,17 @@
+[options]
+mode prove
+depth 8
+
+[engines]
+smtbmc yices
+
+[script]
+read -formal -D PFPU32_FCMP pfpu32_cmp.v
+
+
+prep -top pfpu32_fcmp
+
+[files]
+../../rtl/verilog/pfpu32/pfpu32_cmp.v
+../../rtl/verilog/mor1kx-defines.v
+../../rtl/verilog/mor1kx-sprs.v
--- a/bench/formal/pfpu32_cmp_snan_bug.sby
+++ b/bench/formal/pfpu32_cmp_snan_bug.sby
@@ -1,0 +1,17 @@
+[options]
+mode prove
+depth 8
+
+[engines]
+smtbmc yices
+
+[script]
+read -formal -D PFPU32_FCMP -D PFPU32_FCMP_SNAN_BUG pfpu32_cmp.v
+
+
+prep -top pfpu32_fcmp
+
+[files]
+../../rtl/verilog/pfpu32/pfpu32_cmp.v
+../../rtl/verilog/mor1kx-defines.v
+../../rtl/verilog/mor1kx-sprs.v
--- a/rtl/verilog/pfpu32/pfpu32_cmp.v
+++ b/rtl/verilog/pfpu32/pfpu32_cmp.v
@@ -199,4 +199,203 @@ assign inv_o      = inv_cmp;
 assign inf_o      = infa_i | infb_i;
 assign ready_o    = fpu_op_is_comp_i;
 
+/*-------------------Formal Checking------------------*/
+
+`ifdef FORMAL
+
+`ifdef PFPU32_FCMP
+`define ASSUME assume
+`else
+`define ASSUME assert
+`endif // PFPU32_FCMP
+
+  always @(*) begin
+    // Assume that a valid opcode is given when this unit is activated.
+    `ASSUME (
+      ~fpu_op_is_comp_i
+    | generic_cmp_opc_i == GENERIC_SFEQ
+    | generic_cmp_opc_i == GENERIC_SFNE
+    | generic_cmp_opc_i == GENERIC_SFLT
+    | generic_cmp_opc_i == GENERIC_SFLE
+    | generic_cmp_opc_i == GENERIC_SFGT
+    | generic_cmp_opc_i == GENERIC_SFGE
+    );
+
+    // Assume that inputs are well-formed, based on what is expected from
+    // pfpu32_top.
+    `ASSUME (
+      // Top two bits in first operand's exponent are always 0.
+      exp10a_i[9:8] == 0
+    );
+    `ASSUME (
+      // Top two bits in second operand's exponent are always 0.
+      exp10b_i[9:8] == 0
+    );
+
+    // Assume that inputs are consistent.
+    `ASSUME (
+      // NaN signals for first operand.
+      (snana_i | qnana_i) == (exp10a_i == 255 & fract24a_i != 0)
+    );
+    `ASSUME (
+      // NaN signals for second operand.
+      (snanb_i | qnanb_i) == (exp10b_i == 255 & fract24b_i != 0)
+    );
+    `ASSUME (
+      // Infinity signal for first operand.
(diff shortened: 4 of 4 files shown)


liujed:
As mentioned in my commit message above, I think I might have found a bug. Please let me know how you would like this handled. The fix is a simple one-line (plus comments) change, but I wanted to avoid modifying the actual implementation until the bug is confirmed.

stffrdhrn:
@bandvig on this one could you comment on the bug Jed mentioned?

liujed:
> Also, you mentioned that some of the comparison logic is not described in the spec. Which one? In spec version 1.4 I can see we have defined the lf.sfu* instructions.

I have also been looking at [v1.4](https://raw.githubusercontent.com/openrisc/doc/master/openrisc-arch-1.4-rev0.pdf) of the spec. Here is what I believe is un(der)specified, where it is inconsistent with the implementation, and where it is inconsistent with my understanding of [IEEE 754-2019](https://www-users.cse.umn.edu/~vinals/tspot_files/phys4041/2020/IEEE%20Standard%20754-2019.pdf), §5.11:
> Invalid operation is the only exception that a comparison predicate can signal. All predicates signal the invalid operation exception on signaling NaN operands.

* For = and ≠ (`lf.sfeq.*`, `lf.sfne.*`; pp.151–2, 161–2), the spec says to raise an exception when a signalling NaN is given, but doesn't explicitly specify the behaviour on quiet NaNs. I took qNaNs to never be equal to anything, including itself.
* For unordered-or-= (`lf.sfueq.*`; pp.163–4), the spec says to set the compare flag on any NaN, but the implementation yields an exception on signalling NaNs. IEEE agrees with the implementation, so that is what I went with.
* For unordered-or-≠ (`lf.sfune.d` and `lf.sfune.s`), the spec is missing a behavioural specification, although the machine code reference in §18 lists both instructions. I took these to be the same as `lf.sfne.*`.
* For the unordered-or-inequality comparisons (`lf.sfult.*`, `lf.sfule.*`, `lf.sfugt.*`, `lf.sfuge.*`; pp.165–72), the spec says to set the compare flag on any NaN, but IEEE says to signal an exception for signalling NaNs. This is the suspected bug I mentioned in my commit message.

The ordinary inequality comparisons <, ≤, >, ≥ (`lf.sflt.*`, `lf.sfle.*`, `lf.sfgt.*`) all look fine to me.

liujed:
> @liujed could you rebase this to avoid the conflict that is now reported?

Rebased as 942c8a6. Also removed a couple of unused signals in d808456.

stffrdhrn:
Hiu @liujed 

Bear with me. 

> > Also, you mentioned that some of the comparison logic is not described in the spec. Which one? In spec version 1.4 I can see we have defined the lf.sfu* instructions.
> 
> I have also been looking at [v1.4](https://raw.githubusercontent.com/openrisc/doc/master/openrisc-arch-1.4-rev0.pdf) of the spec. Here is what I believe is un(der)specified, where it is inconsistent with the implementation, and where it is inconsistent with my understanding of [IEEE 754-2019](https://www-users.cse.umn.edu/~vinals/tspot_files/phys4041/2020/IEEE%20Standard%20754-2019.pdf), §5.11:

Thanks, last I looked I was not able to find a copy of IEEE 754, this is helpful.

> > Invalid operation is the only exception that a comparison predicate can signal. All predicates signal the invalid operation exception on signaling NaN operands.
> 
> * For = and ≠ (`lf.sfeq.*`, `lf.sfne.*`; pp.151–2, 161–2), the spec says to raise an exception when a signalling NaN is given, but doesn't explicitly specify the behaviour on quiet NaNs. I took qNaNs to never be equal to anything, including itself.

Right, qNaN is not in spec, but this is the implied behavior as per IEEE 745.  We should clarify this in a future version of the spec.

> * For unordered-or-= (`lf.sfueq.*`; pp.163–4), the spec says to set the compare flag on any NaN, but the implementation yields an exception on signalling NaNs. IEEE agrees with the implementation, so that is what I went with.

OK, I see.

> * For unordered-or-≠ (`lf.sfune.d` and `lf.sfune.s`), the spec is missing a behavioural specification, although the machine code reference in §18 lists both instructions. I took these to be the same as `lf.sfne.*`.

OK, yes, the full spec is missing. Thanks for clarifying.

> * For the unordered-or-inequality comparisons (`lf.sfult.*`, `lf.sfule.*`, `lf.sfugt.*`, `lf.sfuge.*`; pp.165–72), the spec says to set the compare flag on any NaN, but IEEE says to signal an exception for signalling NaNs. This is the suspected bug I mentioned in my commit message.

OK I see.

> The ordinary inequality comparisons <, ≤, >, ≥ (`lf.sflt.*`, `lf.sfle.*`, `lf.sfgt.*`) all look fine to me.

So in summary:

1.  The spec is missing details for how to handle quiet NaN's `lf.sfeq.*`, `lf.sfne.*`
2.  The spec is missing full docs for `lf.sfune.d` and `lf.sfune.s`
3.  The instruction `lf.sfueq.*`, raises an exception for SNaN, but the spec does not say it should.
4.  The inscructions `lf.sfult.*`, `lf.sfule.*`, `lf.sfugt.*`, `lf.sfuge.*`, do not raise exceptions for SNaN which is inline with the spec.  But according to IEEE this should raise an exception.

In other words according to the spec `3` is a bug.  But according to IEEE `4` is a bug and the spec needs updating, to clarify how SNaN should be handled. 

Is this correct?  Let's see what @bandvig has to say.  Also I would like to see how some other architectures handle SNaN for unordered comparisons.

liujed:
For another data point, here is the [AMD64 ISA v4.08](https://www.amd.com/content/dam/amd/en/documents/processor-tech-docs/programmer-references/40332.pdf). They always signal on sNaNs.

From Vol. 1, §6.4.6 (p.357 of the PDF):
> The FUCOMx instructions perform the same operations as the FCOMx instructions, except that the FUCOMx instructions generate an invalid-operation exception (IE) only if an operand is an unsupported data type or a signaling NaN (SNaN), whereas the ordered-compare FCOMx instructions generate an invalid-operation exception if any operand is an unsupported data type or any type of NaN.

stffrdhrn:
> For another data point, here is the [AMD64 ISA v4.08](https://www.amd.com/content/dam/amd/en/documents/processor-tech-docs/programmer-references/40332.pdf). They always signal on sNaNs.
> 
> From Vol. 1, §6.4.6 (p.357 of the PDF):

Thanks, also I checked out QEMU implementation which uses soft float.  It also always signals on sNaN's.  We should update the spec as per your suggestions and fix the bug.

https://gitlab.com/qemu-project/qemu/-/blob/master/fpu/softfloat-parts.c.inc?ref_type=heads#L1631

```c
    if (unlikely(ab_mask & float_cmask_anynan)) {
        if (ab_mask & float_cmask_snan) {. // sNaN  always raises the exception
            float_raise(float_flag_invalid | float_flag_invalid_snan, s);
        } else if (!is_quiet) {.    // unordered operations a quiet, meaning no exception on qNaN
            float_raise(float_flag_invalid, s);
        }
        return float_relation_unordered;
    }
```

stffrdhrn:
I created this proposal to deal with the spec update: https://github.com/openrisc/openrisc.github.io/pull/19

liujed:
Great, thank you!

bandvig:
Hello @liujed and @stffrdhrn 
Unfortunately now days I couldn't pay enough attention to OpenRISC. I could participate in discussions on weekends mostly. If I understand correctly:
 
> 1. The spec is missing details for how to handle quiet NaN's `lf.sfeq.*`, `lf.sfne.*`
> 2. The spec is missing full docs for `lf.sfune.d` and `lf.sfune.s`
> 3. The instruction `lf.sfueq.*`, raises an exception for SNaN, but the spec does not say it should.

Those require corrections in spec only? It was so long time ago I worked on FPU verilog and OpenRISC spec. I just didn't remember IEEE requirements exactly. However I suppose that quiet NaNs should be **quiet** while signaling NaNs should rise exception.

> 4. The inscructions `lf.sfult.*`, `lf.sfule.*`, `lf.sfugt.*`, `lf.sfuge.*`, do not raise exceptions for SNaN which is inline with the spec.  But according to IEEE this should raise an exception.

As I always try to follow IEEE standard it should be corrected of cause.
@liujed If your are ready to propose a fix, let's do it.
@stffrdhrn Will the fix require to update glibc and other libraries ports?

liujed:
Proposed fix is in #167.

PR #166, opened 2025-03-22
Formal verification for pfpu32_muldiv: MULT (and some DIV)

liujed:
This adds formal verification for the multiplication outputs in pfpu32_muldiv, and some division outputs too.

To achieve this, I assumed (but not yet verified) that the muldiv module will not receive further requests for the 12 cycles following a division request, since the module will be busy with the Goldschmidt algorithm.

To make verification of the multiplication output tractable, I've split it into two pieces. The `pfpu_muldiv_check` target in the Makefile checks a series of assertions that together imply an assumption in the `pfpu_muldiv` target. Unfortunately, making the first set of assertions isn't enough to guide the solver towards proving the assumption within a reasonable amount of time.

~I also had to fix a couple of bugs: 5ba34a3.~

### DIFF
--- a/rtl/verilog/pfpu32/pfpu32_muldiv.v
+++ b/rtl/verilog/pfpu32/pfpu32_muldiv.v
@@ -642,6 +642,48 @@ module pfpu32_muldiv
       muldiv_rdy_o <= s2o_mul_ready | s3o_div_ready;
   end // posedge clock
 
+/*-------------------Formal Checking------------------*/
+
+`ifdef FORMAL
+
+`ifdef PFPU32_MULDIV
+`define ASSUME assume
+`else
+`define ASSUME assert
+`endif // PFPU32_MULDIV
+
+  // Reset the module before formal verification.
+  reg f_initialized;
+  initial f_initialized = 1'b0;
+  begin
+    always @(posedge clk)
+      f_initialized <= 1'b1;
+
+    always @(*)
+      if (!f_initialized)
+        assume (rst);
+      else
+        assume (!rst);
+  end
+
+  // Division takes 16 cycles. Verify that a result is produced within this
+  // amount of time.
+  generate
+  begin : f_muldiv_multiclock
+    f_multiclock_pfpu32_op #(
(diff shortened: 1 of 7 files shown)


rtl/verilog/pfpu32/pfpu32_muldiv.v
bandvig, line 1: @liujed 
I'm not sure that there is a bug here. I designed the module many years ago. But, if I remember correctly a next mult/div operation should go through stages just after previous div goes through last pass of Goldshmidt iteration. Perhaps, your changes block the features. I think there are also some other tricks and none obvious behavior of the pipe to speed up its operation but keep results correct. I verified PFPU with https://github.com/bandvig/or1k_testfloat and a FP- benchmark (sorry, not remember which one and no the benchmark sources in my repo).

bandvig (commented):
@liujed
I'm not sure that there is a bug here. I designed the module many years ago. But, if I remember correctly a next mult/div operation should go through stages just after previous div goes through last pass of Goldshmidt iteration. Perhaps, your changes block the features. I think there are also some other tricks and none obvious behavior of the pipe to speed up its operation but keep results correct. I verified PFPU with https://github.com/bandvig/or1k_testfloat and a FP- benchmark (sorry, not remember which one and no the benchmark sources in my repo).

bandvig:
@liujed 
I recall. There is no problem with pfpu32_muldiv's internal pipes control because next command never enters into PFPU32 till completion previous one. This is guaranteed by
https://github.com/openrisc/mor1kx/blob/83d6d49de2c9764e9de8e05f7b897275fe3f9dc2/rtl/verilog/mor1kx_execute_alu.v#L847
In fact, even so the PFPU32 is pipe-lined, the feature is not exploited in CAPPUCCINO. The pipe-lined architecture just brings potentially faster implementation here.
On the other hand, you are right in a case if somebody would reuse the code as a part of more complex design.
So, please re-create your merge request with you fix https://github.com/openrisc/mor1kx/commit/5ba34a32a4811a703f507d2b94e3d21e7cc562b1

PR #167, opened 2025-03-24
Always signal when comparing with signalling NaN

liujed:
Follows up on discussion in #164. Previously, comparisons with signalling NaN would set `FPCSR[IVF]` only for ordered/unordered EQ/NE. Now, all comparisons set this flag when operating on a signalling NaN.

Updated formal verification of `pfpu_cmp` to match.


stffrdhrn:
Thank @liujed for working on this

